PR #1, opened 2021-10-22
Thunder Support

jlacivita:
- Added a new thunder module for converting thunder interface schemas to OpenRPC
- Stopped squashing module names to lower case in RPC calls
- Centralized event name generation into new method in javascript module
- Added safe name creation for JavaScript
- Fixed some minor bugs


scthunderbolt:
Closed due to inactivity, please restore if still needed.

PR #2, opened 2021-10-29
Common definition for OfferingType

onazol:


onazol:
related PR for common definition of OfferingType 
https://github.com/rdkcentral/firebolt-openrpc/pull/2

src/schemas/entertainment.json
```diff
@@ -138,14 +149,7 @@
...
         "offeringType": {
-          "type": "string",
-          "enum": [
-            "free",
-            "subscribe",
-            "buy",
-            "rent"
-          ],
-          "description": "The offering type of the WayToWatch."
+          "$ref": "https://meta.comcast.com/firebolt/entertainment#/definitions/OfferingType"
```
jlacivita, line 152: This is totally a valid JSON $ref value, but could you please change it to be relative (it's referring to a definition in the same file), e.g. `#/definitions/OfferingType` w/out the protocol & domain?

Our tooling assumes that $refs that start with http are from other files, and this might generate strange docs if we do it this way.

jlacivita (changes requested):

src/schemas/entertainment.json
jlacivita: Actually, I guess we're doing this already elsewhere, and it works fine!

jlacivita (dismissed):

jlacivita (approved):
LGTM

PR #5, opened 2021-11-10
Adds option parser

cletustboone:
I hadn't looked at this in a while and it was pretty hard to tell how the cli was being used. There was a hack stuffed in to get a quick feature out of one of the util scripts that made it hard to understand how to use that feature (e.g. `asPath` and `staticModules`).

This makes all the script options explicit. I updated the README with the usage information.

I used the trusty [nopt](https://github.com/npm/nopt) package as the options parser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,48 @@ title: Firebolt OpenRPC Tools
 # Firebolt OpenRPC Tools
 Tools to enable consistent Firebolt SDK generation.
 
-## Usage
-To install the package, run:
+## To Use
 
-```
-npm install @firebolt-js/openrpc
+  - First, `npm install`.
+  - Next, invoke the cli directly with `src/cli.js`. See below for arguments.
+
+## CLI Arguments
+
+`--task`: Tell the tool what you want it to do. `sdk|docs|openrpc|validate`. More on each below.
+
+`--source`: The relative or absolute path to the folder or file that the task will use.
+
+`--template`: The relative or absolute path to the folder or file containing template resources for the task.
+
+`--output`: The relative or absolute path to the folder or file that will hold the task's generated output.
+
+### SDK Generation
+
+Indicated by `--task sdk`. Generate the Firebolt JavaScript SDK from a json-rpc spec.
+
+### Document Generation
+
+Indicated by `--task docs`.
+
+### OpenRPC Generation
+
+Indicated by `--task openrpc`. Reads a spec for a json-rpc API and generates an openrpc document for it.
+
+### JSON-RPC Validation
+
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -2,11 +2,59 @@
...
+## To Use
 
-```
-npm install @firebolt-js/openrpc
-```
\ No newline at end of file
+  - First, `npm install`.
+  - Next, invoke the cli directly with `src/cli.js`. See below for arguments.
+
+Alternatively, you may install globally using `npm install -g @firebolt-js/openrpc`, which will install the cli and add it to your $PATH so it may be invoked with the `firebolt-openrpc` command. However, trying this at the time of writing, it did not work.
```
jlacivita, line 12: Any thoughts on what didn't work?

I tried to model it after standard CLI patterns I found, but I've only been using it via other npm packages, so true CLI wasn't needed. Maybe we should create an issue, and remove this sentence till it's fixed?

README.md
```diff
@@ -2,11 +2,59 @@
...
+  - `openrpc`: Generate the openrpc
+  - `validate` Validate the openrpc
+
+`--source`: The relative or absolute path to the folder or file that the task will use.
+
+`--template`: The relative or absolute path to the folder or file containing template resources for the task.
+
+`--output`: The relative or absolute path to the folder or file that will hold the task's generated output.
+
+`--as-path`: This has some effect on the document generation task. Not sure what.
```
jlacivita, line 29: This is a toggle for generating content as files vs folders, e.g.:

`/docs/Topic.md`

vs

`/docs/Topic/index.md`

It effects output of:

- file names
- relative links
- relative images

jlacivita: When deploying docs to web servers, --as-path is generally used. For deploying to GitHub pages/wikis, do not use --as-path.

README.md
```diff
@@ -2,11 +2,59 @@
...
+
+`--static-modules`: Used by the SDK generator. Makes sure that statically defined modules get exported. An example is `Platform`.
```
jlacivita, line 31: Might be worth mentioning that a statically defined module is one w/out an open-rpc .json document, that is included statically in the template directory. Listing it in this property ensures it will be exported by the SDK, and it will have mock responses wired up as well.

jlacivita (changes requested):

README.md
cletustboone: Good call. Trying again, I'm not totally unsure it wasn't user error on my part. I'll remove the sentence and investigate further. Will create an issue if needed.

README.md
cletustboone: Perfect. Thanks for the clarification. I will update.

jlacivita (approved):
LGTM

PR #6, opened 2021-12-16
Generate code for property methods

kpears201:


src/template/js/sdk/Events/index.js
```diff
@@ -127,11 +127,18 @@ const once = function(...args) {
 }
 
 const listen = function(...args) {
+  init()
   const [module, event, callback] = getListenArgs(...args)
   return doListen(module, event, callback, false)
 }
 
-setListener(listen)
+const init = () => {
```
kpears201, line 135: There was a js module dependency here. setEmitter and setListener are methods used from Transport, but depending on when the modules are loaded, if Events was loaded before Transport, then this doesn't work because the methods are not defined yet. It is usually not a good idea to perform an action when the module is loaded into memory. So I have changed this to lazily set the emitter and listener.

src/template/js/sdk/Transport/mock.js
```diff
@@ -79,11 +79,12 @@ function receive(_callback) {
 }
 
 function event(module, event, value) {
-  const id = Object.entries(event_map).find(([k, v]) => v === module.toLowerCase() + '.' + event.toLowerCase())[0]
-  if (id) {
+  const eventMap = event_map
+  const listener = Object.entries(eventMap).find(([k, v]) => v === module.toLowerCase() + '.' + event)
+  if (listener) {
```
kpears201, line 84: This was a bug, if a listener was never set it would return null before the `[0]` lookup. Changed to first check if the listener exists before accessing the id at `listener[0]`

src/template/js/sdk/Transport/mock.js
```diff
@@ -79,11 +79,12 @@ function receive(_callback) {
 }
 
 function event(module, event, value) {
-  const id = Object.entries(event_map).find(([k, v]) => v === module.toLowerCase() + '.' + event.toLowerCase())[0]
-  if (id) {
+  const eventMap = event_map
```
kpears201, line 82: This fixed a weird scoping problem. event_map is a statically loaded module so it gets stored in this method scope statically.

src/template/js/sdk/Events/index.js
jlacivita: Cool. I struggled with the best way to do that when I wrote it. This seem more robust.

util/sdk/macros/index.mjs
```diff
@@ -389,11 +417,16 @@ function generateMethods(json = {}, onlyEvents = false) {
         params: getMethodSignatureParams(moduleName, methodObj, { isInterface: false })
       }
 
-      const template = getTemplateForMethod(methodObj)
+      let template = getTemplateForMethod(methodObj)
+      if (isPropertyMethod(methodObj)) {
+        template = getTemplate('methods/polymorphic-property.js')
```
jlacivita, line 422: Hmmm. I was hoping to keep template -> tag name 1:1.

I was imagining for this, we'd have:

- methods/property/index.js
- methods/property/immutable.js
- methods/property/readyonly.js

And then update getTemplateForMethod to parse out any `:` and use that as the directory name.

Then put simpler code for each type in each of those individual templates.

The way you have it built with the same template for all three adds unused code (and kilobytes) for each property to the final build. I'd rather have the extra stuff in redudant templates to keep the compiled artifact smaller.

jlacivita (changes requested):

util/sdk/macros/index.mjs
kpears201: It doesn't really add extra stuff because it is meant to throw an error to the client (from the SDK without hitting FBOS) if they try and misuse a property.
Also it is using a reusable module I created `Prop`. So each of these methods just calls into the Prop module with the values of the flags. So it isn't using any more space.

kpears201: the template is this:
```
function ${method.name}() {
  return Prop.prop('${info.title}',  '${method.name}', arguments, ${method.property.immutable}, ${method.property.readonly})
}
```
So basically the smallest amount of code generated per property.
I think it would be a lot more if we had a template per tag type.

jlacivita: yeah, looking more closely I can see that the way you wrote it is smaller (in kb) than how I suggested.

src/template/js/sdk/Transport/mock.js
```diff
@@ -83,7 +83,7 @@ function event(module, event, value) {
   if (id) {
     let message = JSON.stringify({
       jsonrpc: '2.0',
-      id: id,
+      id: listener[0],
```
jlacivita, line 86: I'm confused by this line. Isn't `listener` a reference to a function, via `setListener`

kpears201: merge conflict resolution issue, fixed it.

jlacivita (approved):

PR #8, opened 2022-01-07
Share a transport layer between multiple SDKs

kpears201:


kpears201:
@jlacivita  can you review?

kpears201:
There was some major scoping issues with the transport layer that needed to be addressed to support multiple SDKs.
The transport layer was scoped to a file/module, which means if there was 2 SDKs there were 2 transport layers. This also meant that there was two id counters for the messages.
So if core-SDK sent:
```
{ id: 1, method: "lifecycle.onForeground"}
```
Then manage SDK sent:
```
{ id: 1, method: "appcatalog.catalog" }
```
And FireboltOS responded:
```
{ id: 1, result: { "apps": [] } }
```
How does the SDK know this response is meant for the manageSDK and not the coreSDK?
We could maybe treat the two SDKs as separate clients just as if they were different apps. This would work for the websocket transport because there would be two separate websocket connections and FireboltOS can just respond through the correct one. Unfortunately that does not work with the Legacy transport. We can only target messages to a browser, not a specific client within a browser. 
The solution is to only have a single transport per app, not per SDK, and thus only have a single id counter.
In order to accomplish this, we had to make the transport a global variable so it can be used by both SDKs. This would be very hard with how the transport file was written, as every variable was scoped to the module/file. So I changed the transport file to be an ES5 class instead (this is better practice anyway). And then instantiated that ES5 class as a singleton global variable to be shared across SDKs. One caveat was in the off-device static mock case, we should NOT use the same transport. As the mock transport contains defaults specific to the SDK it is in.
The end changes here look like a lot, but most of it is just moving methods into an ES5 class and change variable scoping to be `this.`. Viewing this review without whitespace changes helps a little bit: https://github.com/rdkcentral/firebolt-openrpc/pull/8/files?w=1

PR #9, opened 2022-01-12
Feature: External Markdown Descriptions

jlacivita:
- Added a descriptions module to load .md files stored in /src/descriptions/
- Updated json-schema module to use descriptions module to inject md
- Using these new features in build:openrpc and build:docs

### DIFF
--- a/util/shared/descriptions.mjs
+++ b/util/shared/descriptions.mjs
@@ -1,0 +1,44 @@
+/*
+ * Copyright 2021 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { fsReadFile, bufferToString } from './helpers.mjs'
+import path from 'path'
+
+const pathDelimiter = '/src/'
+const descriptions = {}
+
+// A through stream that expects a stream of filepaths, reads the contents
+// of any .json files found, and converts them to POJOs
+// DOES NOT DEAL WITH ERRORS
+const loadMarkdownContent = fileStream => fileStream
+    .filter(filepath => path.extname(filepath) === '.md')
+    .flatMap(filepath => {
+        return fsReadFile(filepath)
+            .map(bufferToString)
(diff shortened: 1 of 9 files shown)


util/shared/descriptions.mjs
```diff
@@ -0,0 +1,44 @@
...
+
+// A through stream that expects a stream of filepaths, reads the contents
+// of any .json files found, and converts them to POJOs
+// DOES NOT DEAL WITH ERRORS
+const loadMarkdownContent = fileStream => fileStream
+    .filter(filepath => path.extname(filepath) === '.md')
+    .flatMap(filepath => {
+        return fsReadFile(filepath)
+            .map(bufferToString)
+            .tap(data => descriptions[filepath.split(pathDelimiter)[1]] = data)
```
cletustboone, line 33: Not asking for a change here, but as info, Node has a helpful module for doing this kind of stuff in a platform-agnostic way.

https://nodejs.org/api/path.html

util/shared/json-schema.mjs
```diff
@@ -36,6 +41,70 @@ const getAllSchemas = _ => {
...
+    const field = path.pop()
+
+    // reassign value
+    const node = getPathOr(null, path, obj)
+    node[field] = md
+  })
+
+  return obj
+
+/*
```
cletustboone, line 73: Is there a reason to keep this lying around?

util/shared/descriptions.mjs
jlacivita: Good to know, thanks!

util/shared/json-schema.mjs
jlacivita: Well, it's used in a stream to add all of the markdown content to the obj passed in. if we don't return it, it wouldn't be very streamy. If we take this out, we'd have to change the `.map()` to a `.tap()` in the `getSchemaContent` method.

At the moment it's the final step in the stream, but other steps could be added in the future.

jlacivita: I suppose you could make the case that it should make a copy of the object and do its operations.

util/shared/modules.mjs
```diff
@@ -204,11 +205,11 @@ const getAllModules = _ => {
 // A through stream that expects a stream of filepaths, reads the contents
 // of any .json files found, and converts them to POJOs
 // DOES NOT DEAL WITH ERRORS
-const getModuleContent = fileStream => fileStream
-    .filter(filepath => path.extname(filepath) === '.json')
-    .flatMap(fsReadFile)
-    .map(bufferToString)
-    .map(JSON.parse)
+const getModuleContent = getSchemaContent //fileStream => fileStream
```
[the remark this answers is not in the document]
jlacivita: good catch, thx

cletustboone (approved):
The changes make sense and seem fine to me. I'd get rid of some of the commented out code though.

util/shared/json-schema.mjs
jlacivita: Oh... 😳

Uhm... I wanted "Jeremy was here" in the commit history?

Fixed.

cletustboone (approved):
👍🏻

PR #11, opened 2022-01-19
Document all tags that openrpc uses

kpears201:


README.md
```diff
@@ -77,6 +77,17 @@ It affects the output of:
 
 Indicated by `--task openrpc`. Assembles a corpus of individual OpenRPC documents into a single OpenRPC document.
 
+#### Tags
+* __event__ - When a method is tagged with this then no SDK method is generated. Events are documented in a different section of the module. The app can use the event not through a method call but by doing `Module.listen('eventName')`
```
jlacivita, line 81: it's true that "no SDK method is generated" but I think that's a detail that I wouldn't put in the very first sentence, since other tags also achieve that. Maybe start with

> When a method is tagged with this then it will be treated as an asynchronous event that receives a success response, and then zero or more asynchronous responses with the same ID. These responses correlate to the event happening.

Feel free to wordsmith.

README.md
```diff
@@ -77,6 +77,17 @@ It affects the output of:
...
+* __exclude-from-sdk__ - When a method is tagged with this then no SDK method is generated. Instead, custom code for that method is used. This is often used for methods that have client code associated with it instead of just calling the Transport directly.
+* __polymorphic-pull__ - Instructs the code generation to generate a single method that be used for both pushing data to Firebolt or registering a callback that Firebolt can use to pull data from the application.
+* __polymorphic-reducer__ - Instructs the code generation to generate a single method that can take a single object or an array of objects.
+* __rpc-only__ - No SDK method or docs are generated for this method, but FireboltOS should still handle the RPC call. This is used for internal communication within the SDK to FireboltOS, but is not meant to be consumed by an application.
+* __synchronous__ - Almost all firebolt methods are asynchronous because that make an asynchronous call through the Transport. Some calls which are handled entirely client side can be marked as synchronous and thus do not return a Promise.
```
jlacivita, line 86: "that" -> "they"

README.md
```diff
@@ -77,6 +77,17 @@ It affects the output of:
...
+* __property__ - Generates a single method that can be used as a getter, setter, and subscription based on the arguments the app gives to that method call.
```
jlacivita, line 88: might be worth mentioning the three values of the argument here, so folks know they can do just getters.

README.md
```diff
@@ -77,6 +77,17 @@ It affects the output of:
...
+* __property:readonly__ - Generates a single method that can be used as a getter and subscription based on the arguments the app gives to that method call.
```
jlacivita, line 89: oops, you did what I suggested already

jlacivita (changes requested):

jlacivita (approved):
LGTM

PR #12, opened 2022-01-20
Feature: Deprecated Methods

jlacivita:
Add `deprecated` tag support for methods:

- does a `console.warn()` when calling a deprecated method
- Adds a note to the docs for each one
- Adds an annotation to the declarations file for each one
- Supports optional `x-alternative` attribute in openrpc to enhance the above three features
- Supports an optional `x-since` attribute in openrpc to enhance the above three features

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@firebolt-js/openrpc",
-  "version": "1.2.0-beta.1",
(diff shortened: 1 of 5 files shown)


cletustboone:
Related to https://github.com/rdkcentral/firebolt-core-sdk/pull/9

kpears201:
looks good

PR #13, opened 2022-01-20
Generate OpenRPC (and docs) from implied property events

jlacivita:
This fix generates OpenRPC for the events implied by the `property` and `property:readonly` tags. This results in:

- Documentation for the `<Module>.listen()` API for those events
- Proper event registration of each module, so that `<Module>.listen('*', ...)` will include them
- OpenRPC specs for those events, which power various automation tooling, and serve as a spec for implementation

### DIFF
--- a/util/shared/modules.mjs
+++ b/util/shared/modules.mjs
@@ -202,10 +202,49 @@ const getAllModules = _ => {
     return Object.values(modules)
 }
 
+const createEventFromProperty = property => {
+    const event = JSON.parse(JSON.stringify(property))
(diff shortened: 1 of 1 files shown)


kpears201:
Just those two minor things, rest looks good

jlacivita:
> Think this broke somewhere:

fixed

> Should "Additional Events" show up in the toc?

I'm thinking no because they're sort of implementation details that we support for backwards compatibility. Also, you can click on Events to get to them, and it's non-trivial to add it to the ToC only when they exist.

kpears201 (approved):

PR #15, opened 2022-01-21
fix(documentation): Adding more info from OpenRPC

jlacivita:
- adding documentation for patternProperties schemas
- fixing See Also links to actually work

### DIFF
--- a/util/docs/macros/index.mjs
+++ b/util/docs/macros/index.mjs
@@ -604,6 +604,11 @@ function getExternalSchemaLinks(json) {
 
     const isModule = currentSchema.info
 
+    // Generate list of links to other Firebolt docs
+    //  - get all $ref nodes that point to external files
+    //  - dedupe them
+    //  - convert them to the $ref value (which are paths to other schema files), instead of the path to the ref node itself
+    //  - convert those into markdown links of the form [Schema](Schema#/link/to/element)
     let links = getExternalSchemaPaths(json)
         .map(path => getPathOr(null, path, json))
         .filter(path => seen.hasOwnProperty(path) ? false : (seen[path] = true))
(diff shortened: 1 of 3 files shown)


util/docs/macros/index.mjs
```diff
@@ -605,15 +605,10 @@ function getExternalSchemaLinks(json) {
...
-        .filter(path => path.filter(item => item === 'items').length <= 1) // Schemas burried behind more than one array aren't relevant
         .map(path => getPathOr(null, path, json))
-//        .filter(path => /^modules\//.test(path))
         .filter(path => seen.hasOwnProperty(path) ? false : (seen[path] = true))
         .map(path => _options.baseUrl + getLinkFromRef(path, _options.asPath))
-//        .map(path => path.split('#')[0].split('/').pop().split('.')[0] + 'Schema#' + path.split('/').pop()) // reformat for git wiki
-//        .map(path => path[0].toUpperCase() + path.substr(1)) // initial caps for git wiki
-        //.map(path => (' - [' + path.split('#')[1] + (options.asPath ? '](../' : '](./') + path.split('#')[0] + '#' + path.split('#')[1].toLowerCase() + ')'))
-        .map(path => ' - [' + path.split("/").pop() + '](' + (_options.asPath ? path.toLowerCase() : path) + ')')
+        .map(path => ' - [' + path.split("/").pop() + '](' + (_options.asPath ? path.split('#')[0].toLowerCase() + '#' + path.split('#')[1].split('/').pop().toLowerCase()  : path) + ')')
```
cletustboone, line 611: What is the thing you are after from `path` here? I just spent a few minutes trying to unravel it. Adding a sample input/output as a comment would go a long way towards someone actually understanding this line at a glance.

jlacivita: Generate list of links to other Firebolt docs
  - get all $ref nodes that point to external files
  - dedupe them
  - convert them to the $ref value (which are paths to other schema files), instead of the path to the ref node itself
  - convert those into markdown links of the form [Schema](Schema#/link/to/element)

jlacivita: added comment resembling above

cletustboone (approved):
Thank you for the clarity! 👍🏻

PR #16, opened 2022-01-25
fix(errors): Output pretty JSON parsing errors w/ filename

jlacivita:
We were getting ugly errors w/ no filename on JSON parse errors. Super unhelpful.

This feature adds:

- Code to track which file is being parsed
- Error catching that outputs a more readable error (w/ colors!) and a filename
- Moves individual module parsing _before_ merged OpenRPC parsing, to facilitate more useful debugging
- removes hard coded assumption that validate will always validate the merged file

### DIFF
--- a/util/shared/json-schema.mjs
+++ b/util/shared/json-schema.mjs
@@ -83,9 +83,9 @@ const loadAndParseSchema = filepath => h.of(filepath)
     .map(bufferToString)
     .map(JSON.parse)
     .errors( (err, push) => {
-      err.message = path.basename(filepath) + ": " + err.message
+      err.message = filepath + ": " + err.message
       console.error(`\n\x1b[41m ERROR:\x1b[0m ${err.message}\n`)
-      process.exit(1)
+      push(nil, err)
     })
     .map(addExternalMarkdown)
 
--- a/util/shared/json-schema.mjs
+++ b/util/shared/json-schema.mjs
@@ -76,9 +76,17 @@ const addExternalMarkdown = obj => {
 // DOES NOT DEAL WITH ERRORS
 const getSchemaContent = fileStream => fileStream
     .filter(filepath => path.extname(filepath) === '.json')
+    .flatMap(loadAndParseSchema)
+
+const loadAndParseSchema = filepath => h.of(filepath)
     .flatMap(fsReadFile)
     .map(bufferToString)
     .map(JSON.parse)
+    .errors( (err, push) => {
+      err.message = path.basename(filepath) + ": " + err.message
+      console.error(`\n\x1b[41m ERROR:\x1b[0m ${err.message}\n`)
+      process.exit(1)
+    })
     .map(addExternalMarkdown)
 
 const getRef = (ref, json) => {
--- a/util/validate/index.mjs
+++ b/util/validate/index.mjs
@@ -46,10 +46,6 @@ const run = ({
   const externalFolder = path.join(__dirname, '..', '..', 'src', 'external')
   const modulesFolder = path.join(srcFolderArg, 'modules')
   
-  const load = path => h.of(path)
-          .through(getSchemaContent)
-          .tap(addSchema)
-
   const validate = (json, validator, prefix) => h(validator(json))
           .tap(result => {
             if (result.valid) {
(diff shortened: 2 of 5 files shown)


util/validate/index.mjs
```diff
@@ -45,78 +45,89 @@ const run = ({
...
+      .errors( (error, push) => {
+        console.log(error)
+      })
+      .flatMap(x => {return h(validateOpenRpc(x))})
+      .tap(result => {
+        if (result.valid) {
+          logSuccess(`Module: ${result.title} is valid`)
+        }
+        else {
+          console.error(`\nERROR: ${result.title} failed validation`)
```
cletustboone, line 119: Not really asking for a change here, but highland has native support for handling errors in the stream. I can mark this down for a refactor later.

util/validate/index.mjs
```diff
@@ -45,78 +45,89 @@ const run = ({
...
+      .through(getSchemaContent)
+      .tap(addSchema)
+      .collect()
+      .flatMap(_ => recursiveFileDirectoryList(schemasFolder).flatFilter(isFile))
+      .through(getSchemaContent)
+      .tap(addSchema)
+      .collect()
+      // Validate all of the JSON-Schemas
+      .flatMap(_ => h(getAllSchemas()))
+      .flatMap(x => {return h(validateJsonSchema(x))})
```
cletustboone, line 81: Sorry for the annoying "style" comment, but this can be a one-liner like the other simple `.flatMap` s in here:
```js
.flatMap(x => h(validateJsonSchema(x))
```

util/validate/index.mjs
cletustboone: Also, this is the same logic as lines 97-105, so this can be consolidated into a single function.

cletustboone:
Is this a bug fix or a feature? Smells like a feature to me.

util/validate/index.mjs
jlacivita: thanks

util/validate/index.mjs
jlacivita: not annoying at all!

jlacivita:
I had the same question... but I leaned towards bug, because the error messages existed, they were just lousy. Open to discussing, so we can start to form a consensus on what's what.

cletustboone:
Fix works for me, actually. In the tooling, maybe it's a bit of a grey area.

util/shared/json-schema.mjs
```diff
@@ -76,9 +76,17 @@ const addExternalMarkdown = obj => {
 // DOES NOT DEAL WITH ERRORS
 const getSchemaContent = fileStream => fileStream
     .filter(filepath => path.extname(filepath) === '.json')
+    .flatMap(loadAndParseSchema)
+
+const loadAndParseSchema = filepath => h.of(filepath)
     .flatMap(fsReadFile)
     .map(bufferToString)
     .map(JSON.parse)
```
cletustboone, line 84: Okay, so if `JSON.parse` throws an error, you deal with it in the handler and terminate the process?

cletustboone: You can optionally just do the `console.error` and `push(null, {})` and let the program keep running. Or do you want a CI process to fail on a single error?

jlacivita: Was going for a single error and exit, but now that they are pretty i think you are right.

util/validate/index.mjs
```diff
@@ -45,78 +45,71 @@ const run = ({
...
   const externalFolder = path.join(__dirname, '..', '..', 'src', 'external')
   const modulesFolder = path.join(srcFolderArg, 'modules')
-  const distFolder = path.join('dist')
   
-  // Load all of the JSON-Schemas
+  const load = path => h.of(path)
+          .through(getSchemaContent)
+          .tap(addSchema)
+
+  const validate = (json, validator, prefix) => h(validator(json))
```
cletustboone, line 53: Are any of the arguments to this function asynchronous? I can't figure out where `validator` comes from.

jlacivita: validator is async, yes. it comes from the stream code in each of the if/else

util/shared/json-schema.mjs
jlacivita: Okay, I bubbled up the errors, and now list all files with errors in them.

util/validate/index.mjs
cletustboone: Yeah, I looked at it for a minute, but didn't immediately spot anything that was actually asynchronous. I see that you put `async` in front of it, but not sure why. Can revisit later. What I was getting at is that you have a `.flatMap(...).flatMap(...)` later in the code that maybe only needs to be `.flatMap(...).map(...)`. In other words you are doing async when you don't need to, maybe.

cletustboone (approved):
Looks good 👍🏻

PR #17, opened 2022-01-25
test(Jest): Configures Jest to work properly with ESM Modules

cletustboone:
## Motivation

I'm operating under the assumption that we want to write tests for the code that generates the docs, the sdk, etc. All of that code lives in these `.mjs` files as ESM modules. Jest only supports ESM experimentally without help from babel. I wanted to avoid babel altogether to start with (more on that below), so I followed [Jest's guide to configuring support for ESM modules](https://jestjs.io/docs/ecmascript-modules).

I added a couple of tests to give the `npm test` script something to do. And also as a starting point to illustrate that some of this code, as written, is untestable. But let's save that for after some very basic decisions about testing are made.

## Issues

There is a bunch of [template code mixed into this repo](https://github.com/rdkcentral/firebolt-openrpc/tree/main/src/template/js). Do we also want to test that code? I assume so. If that's the case, it might better belong in `firebolt-core-sdk`. If we keep it here, because the _template code_ targets browsers, we must setup 2 jest configs.

Now, for code targeting browsers, [Jest can use Babel](https://jestjs.io/docs/getting-started#using-babel) or [webpack](https://jestjs.io/docs/webpack) to transpile the ESMs to code understood by a browser. I suggest we pick one and stick with it in both this repo and `firebolt-core-sdk`.

By default, Jest will look for tests in a folder called `__tests__` that lives next to your code. Like this:

```
util/
├── shared
│   ├── __tests__
│   │   └── descriptions.test.mjs
│   ├── descriptions.mjs
```

It also has some file patterns it will try to match. A file called `foo.(spec|test).js` will be recognized as a test file as long as it's somewhere in your source files.

I would suggest sticking with the defaults, unless there is a compelling reason not to. Requires less configuration, and hopefully meets our needs here.

## Decisions to be Made

- Do we want to test the doc generation code?
- Are we cool with Jest's defaults for where tests go?
- Do we want to keep the template code here and test it here or move it to the core sdk and test it there?
- Should we pick webpack or babel for packaging code for browser tests?


package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@firebolt-js/openrpc",
+  "type": "module",
```
cletustboone, line 3: ## This needs attention!

Jest needed me to change the `type` property in the `package.json` to `module`. The Jest docs [_saaaay_](https://jestjs.io/docs/ecmascript-modules) that they attempt to detect ECMAScript Modules that same way that NodeJS does, but that isn't true. 

> _Beyond that, we attempt to follow node's logic for activating "ESM mode" (such as looking at type in package.json or .mjs files)_

Jest requires both `.mjs` extension **AND** the `package.json` change to activate ESM mode.

[NodeJS](https://nodejs.org/api/esm.html#esm_enabling), on the other hand, will OR the following:

- Presence of the `.mjs` file extension
- `"type": "module"` in `package.json`
- `--input-type="module"` runtime flag

util/shared/descriptions.test.mjs
```diff
@@ -16,18 +16,16 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-/**
- * This module sets up the mock transport layer immediately, instead of letting the SDK wait 500ms
- */
+import { getMarkdown, getAllMarkdownNames } from './descriptions.mjs'
```
cletustboone, line 19: ## And here are some tests

The presence of these will allow `npm test` to have something to do.

util/shared/descriptions.test.mjs
```diff
@@ -16,18 +16,16 @@
...
- // wires up the mock transport w/out waiting
- win.__firebolt.mockTransportLayer = true
- 
- // sets a flag that mock defaults impls can use to speed things up, e.g. Lifecycle/defaults.js
- win.__firebolt.automation = true
\ No newline at end of file
+test('and by extension, so is getAllMarkdowns', () => {
+  // Yay, we have tested that Object.keys works as described
+  expect(getAllMarkdownNames('blah')).toStrictEqual([])
+})
```
cletustboone: ## And here are some tests

The presence of these will allow `npm test` to have something to do.

package.json
jlacivita: I tried this a long time ago and it caused problems I can't recall... things have evolved quite a bit since then, so I think your "This needs attention" is still true :)

cletustboone:
Closing to re-address after #20 is merged.

PR #19, opened 2022-02-10
Stop smashing module names in OpenRPC to lowercase

jlacivita:
This enables Mocks, reference apps, etc. to know the case-sensitive module names in each SDK.


cletustboone (approved):
Just ran across this in #20 and wondered if it was intentional.

cletustboone:
This change also exists in #20, so this can be closed, I think.

cletustboone:
To be addressed as part of a future non-patch release.

PR #20, opened 2022-02-15
refactor(functional purity): evaluation of shared functions should have no side-effects

cletustboone:
Undoing the pattern of:
```
// some-module.mjs
let foo = {}

let myFunc = bar => {
  // Perform side-effects on foo
}

export {
  myFunc
}
```
Then elsewhere...
```
// elsewhere-1.mjs
import { myFunc } = './some-module.mjs'

// Performs some side-effects on `foo` in the context of `bar`
myFunc("bar")
```
and
```
// elsewhere-2.mjs
import { myFunc } = './some-module.mjs'

// Performs additional side-effects on `foo` in the context of `baz`
myFunc("baz")
```
## But Why?

This probably seems overly academic. But I'll attempt to demonstrate why this matters.

The above pattern makes it so that any code anywhere can modify state at any time in a shared global context. Maybe that doesn't seem so bad, but when you go to write a unit test for a function that reads or writes to something in that shared global context, you really have no idea what assumptions your function is making about the shared state upon which it's operating because it could have been modified by some other part of the program without your knowledge. The only way to meaningfully test this scenario is to essentially rebuild your entire application state in your test harness. And at that point, you don't really have a unit anymore. You just have your application again. You can still write tests that validate the expected _side-effects_ of your code. Sometimes called "integration tests" or "functional tests," with the code in this state, any time those tests break, tracing the problem will be extraordinarily difficult because the tests still don't show you how a side-effect came to exist.

## The Root of the Issue: Bringing Together Sync and Async Context

When working with streams processing, getting all of your state in the right place can be a head-spinning proposition. Luckily, there are concepts from functional programming that can help us here. When you need to do something like iterate over a collection, but you need context from an operation further upstream. For example, let's say you aynchronously computed a `version` to be used in each item in the collection. But how to get that `version` into the context of an async function further down the stream? It can be extremely tempting to do something like this:
```js
const version = loadVersion(someFilePath) // This defines a stream
const myCollection = loadCollection() // Another stream

// Initialize a variable outside of stream processing context
const v = {
  foo: ''
}
// Now let's do stuff with the stream
version
  .tap(versionObj => v.foo = versionObj.foo) // <-- perform side effects to variable outside stream processing context
  .flatMap(v => myCollection.map(item => item.v = v)) // <-- use `v` context to do stuff with the collection item :( 
```
What's so wrong with this? Now the closure inside the `myCollection.map` is impure. It uses data, `v`, from outside of its current scope. Now imagine what happened if `v` was in the module's global scope. What if `v` were in a _different_ module's global scope? And now imagine that yet another module relies on `v` having been modified from module A _before_ it was modified from a different function in module B? Imagine writing a test to verify this behavior.

If an alternate approach from the one above looks obvious to you, good! If not, it looks like this:
```js
const version = loadVersion(someFilePath) // This defines a stream
const myCollection = loadCollection() // Another stream

// Now let's do stuff with the stream
version
  .flatMap(versionObj => myCollection.map(item => item.v = versionObj)) // <-- We have `versionObj` in scope from the version.flatMap call, so use it here.

```
But what does it look like to scale this pattern? It seems obvious when you have just one piece of data. What do you do when your processing function needs lots more disparate async context? Well, it's streams all the way down, luckily:
```js
const version = loadVersion(someFilePath) // This defines a stream
const collA = loadCollection('a') // Another stream
const collB = loadCollection('b') // Another
const collC = loadCollection('c') // And another

// Processing function that needs all context:
const processingFunction = version => (collections = []) => {
  // Do stuff with version and collections
  return {foo: 'bar'}
}
```
So we've got a processing function that needs 2 pieces of context. Both are loaded asynchronously via a stream. How do we get all the async context into the synchronous function?
```js
version
  .map(version => processingFunction(version)) // <-- Calling our processing function with the version argument. Gives us back a function that takes the collections as an array.
  .flatMap(fnWithVersion => collA              // <-- Inside the flatMap, we have the partially applied processing function. Let's load our next piece of async context.
    .flatMap(collectionA => collB              // <-- collectionA is now in scope
      .flatMap(collectionB => collC            // < -- collectionB now in scope
        .map(collectionC => fnWithVersion([collectionA, collectionB, collectionC]))))) // <-- All collections and partially applied function in scope. Call it with remaining arguments.
```
There's a lot going on. Let's break it down:

1. First we call our processing function with the first of its arguments, `version`.
2. Once we have that, we need to get the rest of our async data in scope, so we will return a stream, so `.flatMap(fnWithVersion => /* will return a stream */)`
3. The stream we'll return is `collA`. So we do `.flatMap(fnWithVersion => collA)`
4. And since `collA` is a stream, we can `flatMap` over it and return another stream of `collB` and then do the same with `collC`.
5. At last, we have all of our context in scope and we can call our partially applied `fnWithVersion` with the rest of its arguments.
      1. We use `.map(collectionC => ...)` because `processingFunction` returns an object, not another stream. If it returned a stream, the `map` would become `flatMap`.

Knowing all this, there's an even shorter way to write the above:
```js
version.map(processingFunction) // <-- no need to write out the entire arrow function
  .flatMap(fnWithVersion => collA
    .flatMap(collectionA => collB
      .flatMap(collectionB => collC
        .map(collectionC => fnWithVersion([collectionA, collectionB, collectionC])))))
```
## Lifting Functions Into the Context of Their Data

This whole song and dance is known as a "lift." Crocks has functions like `liftA2` or `liftA4`. These are just functions that perform some of the work of lifting functions into contexts. I've found it generally helpful to think of functions and their data this way. For example, in the document generation script, we lift the doc-generation function into the context of 4 pieces of data. This is a `liftA4` that we are implementing with highland streams. The refactored utility scripts utilize this pattern.


jlacivita (approved):
I just ran a diff against our latest stable release in npm and it's identical.

PR #21, opened 2022-02-17
fix(declarations): Create proper declarations for polymorphic pull & properties

jlacivita:
This fix does the following:

- Generates proper TS declarations for `property` & `property:readonly` subscribe APIs
- Generates proper TS declarations for `property` setter APIs
- Generates proper TS declarations for `polymorphic-pull` pull registration APIs
- Hides `roc-only` APIs from TS declarations

### DIFF
--- a/util/declarations/generator/index.mjs
+++ b/util/declarations/generator/index.mjs
@@ -29,7 +29,7 @@ import isString from 'crocks/core/isString.js'
 import predicates from 'crocks/predicates/index.js'
 const { isObject, isArray, propEq, pathSatisfies } = predicates
 
-import { getMethods, getTypes, isEventMethod, isPublicEventMethod, getEnums } from '../../shared/modules.mjs'
+import { getMethods, getTypes, isEventMethod, isPublicEventMethod, isPolymorphicPullMethod, getEnums, isRPCOnlyMethod } from '../../shared/modules.mjs'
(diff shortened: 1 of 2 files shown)


util/shared/modules.mjs
```diff
@@ -111,6 +111,13 @@ const isEventMethod = compose(
     getPath(['tags'])
 )
 
+const isPolymorphicPullMethod = compose(
```
cletustboone, line 114: Hurray for pure functions!

util/declarations/generator/index.mjs
```diff
@@ -29,7 +29,7 @@ import isString from 'crocks/core/isString.js'
 import predicates from 'crocks/predicates/index.js'
 const { isObject, isArray, propEq, pathSatisfies } = predicates
 
-import { getMethods, getTypes, isEventMethod, isPublicEventMethod, getEnums } from '../../shared/modules.mjs'
+import { getMethods, getTypes, isEventMethod, isPublicEventMethod, isPolymorphicPullMethod, getEnums, isRPCOnlyMethod } from '../../shared/modules.mjs'
```
onazol, line 32: Doesn't seem like you are using isPolymorphicPullMethod. Did you mean to import it? ( I could be missing something )

onazol:
ran locally, seeing the missing declarations for entityInfo and purchaseContent as expected

cletustboone (approved):
I don't see any globally modified state or impure functions, which means that the new code you've added _should_ be testable later, which is all I care about right now.

PR #23, opened 2022-03-23
Adding code for generating provider code

kpears201:


util/sdk/macros/index.mjs
```diff
@@ -73,11 +75,32 @@ const isPublicEventMethod = and(
...
+          propEq('name', 'event'),
+          propSatisfies('x-provides', isEmpty)
+        )
+      )
+    ),
     getPath(['tags'])
   )
 )
 
+const isProviderMethod = compose(
```
cletustboone, line 90: Seems like you're also doing this in `util/docs/macros/index.mjs`. Is determining if a method is a provider method or not done differently in the docs macros?

kpears201: It is the same, I don't know why sdk uses crocks and docs doesn't, but I just kept it in the same pattern as what it was already doing with events

jlacivita: cause my small brain was incapable of following @cletustboone's example of crocking & highlighting at the same time 😂

src/template/markdown/examples/provider.md
```diff
@@ -0,0 +1,10 @@
+import { ${module} } from '${package.name}'
+
+class MyProvider {
+  async ${example.providerMethod}(${method.result.name}, responder) {
```
jlacivita, line 4: I'm wondering if we want to find a way to generate one example for all of the methods in the capability's provider interface? otherwise they look optional by only having one at a time.

util/docs/macros/index.mjs
```diff
@@ -30,6 +30,10 @@ import safe from 'crocks/Maybe/safe.js'
 
 var pkg = JSON.parse(fs.readFileSync('./package.json', 'utf8'));
 
+
+const PROVIDER_PREF = "onRequest"
+const PROVIDER_PREF_LEN = "onRequest".length
```
jlacivita, line 35: Should probably change this to `PROVIDER_PREF.length`

src/template/js/sdk/Events/index.js
```diff
@@ -129,6 +140,34 @@ const listen = function(...args) {
   return doListen(module, event, callback, false)
 }
 
+const provide = function(...args) {
```
jlacivita, line 143: Hmm... I was assuming the generic `provide` method would live in a new Capabilities module. We're only using events to implement this cause of our desire to use RPC and only have one client and one server. Seems irrelevant to the App.

Can we move this to a new Capabilities.provide method, w/out causing a headache? It would be a static module, I think. No need for a modules/capabilities.json, it'd just be in src/template/js/sdk/Capabilities/index.js

jlacivita (changes requested):

src/template/markdown/examples/provider.md
jlacivita: Let's get this committed to unblock testing, and we can look into docs later.

PR #25, opened 2022-04-13
Fix Event API return types, support additionalProperties schemas

jlacivita:
This PR fixes a few errors in our TypeScript Declarations:

- adds return types for the Event APIs in each module (listen, once, clear, subscribe)
- removes `patternProperties` from all OpenRPC schemas, as it does not translate to TypeScript
- uses `additionalProperties` as a schema instead (done for `BooleanMap` and `FlatMap`)
- changes event listener callbacks to return `void`, not `{}`

Related PR:
https://github.com/rdkcentral/firebolt-schemas/pull/6

### DIFF
--- a/util/declarations/generator/index.mjs
+++ b/util/declarations/generator/index.mjs
@@ -116,13 +116,13 @@ const generateListeners = (json, schemas = {}) => compose(
   * Listen to all ${getModuleName(json)} events.
   * @param {Function} listener The listener function to handle the events.
   */
-  function listen(listener: (event: string, data: object) => {}): Promise<bigint>
+  function listen(listener: (event: string, data: object) => void): Promise<bigint>
 
   /**
   * Listen to one and only one instance of any ${getModuleName(json)} event (whichever is first).
   * @param {Function} listener The listener function to handle the events.
   */
-  function once(listener: (event: string, data: object) => {}): Promise<bigint>
+  function once(listener: (event: string, data: object) => void): Promise<bigint>
 `
   }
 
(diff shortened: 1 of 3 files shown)


util/docs/macros/index.mjs
```diff
@@ -79,9 +79,9 @@ export {
 function insertMacros(data = '', moduleJson = {}, templates = {}, schemas = {}, options = {}, version = {}) {
     let match, regex
     const methods = moduleJson.methods && moduleJson.methods.filter( method => !isEvent(method) && !isSetter(method) || (isEvent(method) && isFullyDocumentedEvent(method)) || isProviderMethod(method))
-    const additionalEvents = moduleJson.methods && moduleJson.methods.filter( method => isEvent(method) && !isFullyDocumentedEvent(method) && !isProviderMethod(m))
```
cletustboone, line 82: Was this `m` instead of `method` thing just an existing bug?

util/declarations/generator/index.mjs
```diff
@@ -116,13 +116,13 @@ const generateListeners = (json, schemas = {}) => compose(
   * Listen to all ${getModuleName(json)} events.
   * @param {Function} listener The listener function to handle the events.
   */
-  function listen(listener: (event: string, data: object) => {})
+  function listen(listener: (event: string, data: object) => void): Promise<bigint>
```
cletustboone, line 119: Seems like there were two problems here:

1. Callback is required to return an empty object `{}` or `null`.
2. The return type of `listen` wasn't specified at all and is now `Promise<bigint>`.

Is that right?

util/docs/macros/index.mjs
jlacivita: yeah, not sure where it came from, but it broke stuff...

kpears201 (approved):

PR #27, opened 2022-04-28
Generate Beautiful Errors

jlacivita:
- Transform AJV errors into MUCH more human readable output
- Use color escape codes for error readability
- Add example errors to aggregate error list, and let calling method determine what to do (instead of logging an easy to ignore warning)
- call process.exit if there are any errors, but finish validating all files first (this prevents generating docs, sdk, etc., when there are errors)
- added explicit requirement for examples to exist

### DIFF
--- a/util/docs/macros/index.mjs
+++ b/util/docs/macros/index.mjs
@@ -418,6 +418,7 @@ function generatePropertySignatures (m) {
 function iterateSignatures(data, method, moduleJson = {}, schemas = {}, templates = {}, options = {}) {
     // we're hacking the schema here... make a copy!
     method = JSON.parse(JSON.stringify(method))
+    const module = moduleJson.info.title
     let signatures = [method]
     if (hasTag(method, 'property') || hasTag(method, 'property:readonly') || hasTag(method, 'property:immutable')) {
         signatures = generatePropertySignatures(method)
@@ -789,7 +790,7 @@ function generateRPCExample(example, m, moduleJson = {}) {
         return generatePropertyChangedRPCExample(example, m, moduleJson)
     }
     else if (m.tags && m.tags.filter(t => (t.name === 'property-set')).length) {
-        return generatePropertySetRPCExample(example, m, moduleJson)
+        return generatePropertySetRPCExample(example, m, moduleJson.info.title)
     }
     let request = {
         "jsonrpc": "2.0",
--- a/util/docs/macros/index.mjs
+++ b/util/docs/macros/index.mjs
@@ -418,7 +418,6 @@ function generatePropertySignatures (m) {
 function iterateSignatures(data, method, moduleJson = {}, schemas = {}, templates = {}, options = {}) {
     // we're hacking the schema here... make a copy!
     method = JSON.parse(JSON.stringify(method))
-    const module = moduleJson.info.title
     let signatures = [method]
     if (hasTag(method, 'property') || hasTag(method, 'property:readonly') || hasTag(method, 'property:immutable')) {
         signatures = generatePropertySignatures(method)
@@ -482,7 +481,7 @@ function iterateSignatures(data, method, moduleJson = {}, schemas = {}, template
             method.result.schema = possibleResults.find(s => s['$ref'] !== "https://meta.comcast.com/firebolt/types#/definitions/ListenResponse")
         }
         else {
-            console.log(`\nERROR: ${getTitle(module)}.${method.name} does not have two return types: both 'ListenResponse' and an event-specific payload\n`)
+            console.log(`\nERROR: ${getTitle(moduleJson.info.title)}.${method.name} does not have two return types: both 'ListenResponse' and an event-specific payload\n`)
             process.exit(1)
         }
     }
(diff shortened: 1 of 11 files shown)


util/shared/helpers.mjs
```diff
@@ -222,6 +223,8 @@ const localModules = (modulesFolder = '', markdownFolder = '', disableTransforms
     .reduce({}, fileCollectionReducer('/modules/'))
   }
 
+  const trimPath = file => file.startsWith(process.cwd()) ? file.substr(process.cwd().length + 1) : file
```
cletustboone, line 226: It looks like maybe you're trying to trim the working directory off the front of the absolute file path. You can do that with node's `path` utility. For example
```js
const path = require('path');
const file = process.cwd() + '/some/path/to/file.js'
path.relative(process.cwd(), file) // some/path/to/file.js
```

util/validate/index.mjs
```diff
@@ -89,9 +89,16 @@ const run = ({
...
       logSuccess(`${moduleType}: ${result.title} is valid`)
     } else {
-      errorCounter++
-      console.error(result)
-      console.error(`\nERROR, ${moduleType}: ${result.title} failed validation. There are ${errorCounter} other errors`)
+      filesWithErrorCount++
+
+      logError(`${moduleType}: ${result.title} failed validation with ${result.errors.length} errors:`)
+
+      // blank line for readablility
```
cletustboone, line 96: I mean, why not just stick a `\n` on the end of the previous `logError` call?

util/validate/index.mjs
```diff
@@ -89,9 +89,16 @@ const run = ({
     if (result.valid) {
       logSuccess(`${moduleType}: ${result.title} is valid`)
     } else {
-      errorCounter++
-      console.error(result)
-      console.error(`\nERROR, ${moduleType}: ${result.title} failed validation. There are ${errorCounter} other errors`)
+      filesWithErrorCount++
```
cletustboone, line 92: Just to note, there is explicit side-effect stuff going on here.

util/validate/validation/index.mjs
```diff
@@ -19,10 +19,85 @@ import { localizeDependencies } from '../../shared/json-schema.mjs'
 import crocks from 'crocks'
 const { getPathOr } = crocks
 
+const addPrettyPath = (error, json) => {
+  const path = []
+  const root = json.title || json.info.title
+
+  let pointer = json
```
cletustboone, line 26: Why?

util/docs/macros/index.mjs
```diff
@@ -418,6 +418,7 @@ function generatePropertySignatures (m) {
 function iterateSignatures(data, method, moduleJson = {}, schemas = {}, templates = {}, options = {}) {
     // we're hacking the schema here... make a copy!
     method = JSON.parse(JSON.stringify(method))
+    const module = moduleJson.info.title
```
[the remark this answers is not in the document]
jlacivita: Definitely an error that's been hanging out in the code since before this PR. I agree that the usage in 485 is puzzling.

util/validate/validation/index.mjs
jlacivita: The code doesn't mutate the `json` variable or any of it's sub-object, either through the `json` reference nor the `pointer` reference.

The code is using `pointer` to walk through the JSON tree w/out loosing the original reference to the top of the tree.

You could argue that `pointer` isn't needed for this, but it allows for each code editing, for example if we wanted to add `json.info.title` to the returned string, we'd still have access to that part of the object. Seems iffy to reuse a passed in parameter variable to do something like this.

util/docs/macros/index.mjs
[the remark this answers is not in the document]
jlacivita: will remove this variable, and use `moduleJson.info.title` on 485

util/validate/index.mjs
jlacivita: I gotta say, this seems equally bad for different reasons... Code is adding to the error count in three different places now.

https://github.com/rdkcentral/firebolt-openrpc/pull/29

kpears201 (approved):

util/validate/validation/index.mjs
cletustboone: Not quite sure what you mean about not losing the original reference to the top of the tree.

You are mutating `pointer`, and `json` (by reference) [right here](https://github.com/rdkcentral/firebolt-openrpc/blob/b37fdf1370ad481cae5c67c1b5490ab83d64a073/util/validate/validation/index.mjs#L30).

cletustboone (approved):

util/validate/validation/index.mjs
jlacivita: nope, we're just changing what pointer points to. For example:


```javascript
> let json = {foo: {bar: { str: 'hello'}}}
undefined
> json
{ foo: { bar: { str: 'hello' } } }
> let pointer = json
undefined
> pointer
{ foo: { bar: { str: 'hello' } } }
> let pointer = json['foo']
Uncaught SyntaxError: Identifier 'pointer' has already been declared
> pointer = json['foo']
{ bar: { str: 'hello' } }
> pointer
{ bar: { str: 'hello' } }
> pointer = pointer['bar']
{ str: 'hello' }
> pointer
{ str: 'hello' }
> pointer = pointer['str']
'hello'
> pointer
'hello'
> json
{ foo: { bar: { str: 'hello' } } }
> 
```

^^ from node

PR #29, opened 2022-05-06
Make `printResult` functionally pure

jlacivita:
Remove the functionally impure `errorCount++` in printResult

### DIFF
--- a/util/validate/index.mjs
+++ b/util/validate/index.mjs
@@ -56,7 +56,7 @@ const run = ({
   // Set up the ajv instance
   const ajv = new Ajv()
   addFormats(ajv)
-  let errorCounter = 0
+  let filesWithErrorCount = 0
 
   const combinedSchemas = combineStreamObjects(schemaFetcher(sharedSchemasFolder), schemaFetcher(schemasFolder), schemaFetcher(externalFolder))
   const allModules = localModules(modulesFolder, markdownFolder, disableTransforms, false) // Validate private modules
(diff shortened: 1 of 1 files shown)


util/validate/index.mjs
```diff
@@ -133,11 +130,11 @@ const run = ({
...
           .concat(openRpc(jsonSchemaSpec)
             .flatMap(orSpec => validateModules(ajvPackage(ajv, orSpec))(schemas)))))) // ...module validation
-            .collect()
-            .tap(_ => {
-              if (filesWithErrorCount > 0) {
-                // new line for easy-reading
-                console.log()
+            .filter( result => !result.valid ) // check if any results are not valid
+            .collect() // collect them into an array
+            .tap(invalidResults => {
```
cletustboone, line 135: I think you can just `.map` and exit here like so:
```js
.flatMap(validateStuff)
.filter(result => !result.valid)
.collect()
.flatten()
.map(invalidThing => process.exit(-1))
```
This should still achieve the effect of running validation on everything, then exiting if any items make it through the filter. Still exits if _any_ results are invalid.

cletustboone: Also, no need to get super fancy. This makes enough sense for someone to easily understand.

cletustboone (approved):
Looks good.

util/validate/index.mjs
jlacivita: cool. I also added a console.log of the error count, so the array code works better for that.

PR #30, opened 2022-05-06
Wire-up `x-response` for Provider APIs

jlacivita:
This PR adds the following:

- generates `<method>Response` based on presence of an `x-response` schema
- generates `<method>Focus` based on presence of an `x-allow-focus` boolean
- changes the `x-provides` tag to be a single string, rather than an array of strings
- Correlate `provide()` calls to capabilities, not module names
- Standardized provider signature to `<method>(parameters?: Object, session?: Object): Promise<schema>`
- `session` contains correlationId for debugging and optional `focus()` method if the provider has `x-allow-focus`
- support more than one provided capability per module
- support capabilities with more than one method
- updates the generated provider pattern javascript template
- added Provider unit tests
- Fixed a nasty `localizeDependecies` side effect
- Fixed `definitions.<x>` examples validation

Related PRs:
https://github.com/rdkcentral/firebolt-manage-sdk/pull/10
https://github.com/rdkcentral/firebolt-schemas/pull/7


util/shared/modules.mjs
```diff
@@ -303,6 +336,108 @@ const createSetterFromProperty = property => {
...
+        }
+
+        response.result = {
+            name: 'result',
+            schema: {
+                type: 'null'
+            }
+        }
+
+        let n = 1
```
kpears201, line 419: doesn't seem like n is incremented anywhere?

util/shared/modules.mjs
```diff
@@ -303,6 +336,108 @@ const createSetterFromProperty = property => {
...
+        return ready
+    }
+    else {
+        throw "Methods with the `x-provider` tag extension MUST start with 'onRequest'."
+    }
+}
+
+const createResponseFromProvider = (provider, json) => {
+
+    if (provider.name.startsWith('onRequest')) {
```
kpears201, line 381: minor syntax improvement:
```
if (provider.name.startsWith('onRequest')) {
   throw "Methods with the `x-provider` tag extension MUST start with 'onRequest'."
}
// function code
```
That way the entire function doesn't need to be indented

kpears201: Same comment for `createReadyFromProvider`

src/template/js/sdk/Capabilities/index.mjs
```diff
@@ -18,43 +18,82 @@
...
+  const iface = providerInterfaces[capability]
+
+  if (provider.constructor.name !== 'Object') {
+    methods.push(...Object.getOwnPropertyNames(Object.getPrototypeOf(provider)).filter(item => typeof provider[item] === 'function' && item !== 'constructor'))
+  }
+  else {
+    methods.push(...Object.getOwnPropertyNames(provider).filter(item => typeof provider[item] === 'function'))
+  }
+
+  if (iface) {
```
kpears201, line 44: minor syntax improvement:
```
if (!iface) {
    throw "Ignoring unknown provider capability."
}
const valid = iface.every(method => methods.find(m => m === method.name.split('.').pop()))
if (!valid) {
        throw `Provider that doesn not fully implement ${capability}:\n\t${iface.map(m=>m.name.split('.').pop()).join('\n\t')}`
}
if (!defined) {
  return
}
// rest of function
```
Would improve the amount of indentation in this method

src/template/js/sdk/Capabilities/index.mjs
```diff
@@ -18,43 +18,82 @@
...
-        Transport.send(module, methods[i] + 'Ready', {
-          correlationId: req.correlationId
-        })
+        }
       })
-    } else {
-      console.warn("Ignoring unknown provider method '" + module + '.' + methods[i] + "'")
     }
+    else {
+      throw `Provider that doesn not fully implement ${capability}:\n\t${iface.map(m=>m.name.split('.').pop()).join('\n\t')}`
```
kpears201, line 92: typo in error msg

test/suite/provider.with-multiple-methods.test.js
```diff
@@ -0,0 +1,148 @@
...
+let providerMethodTwoResultSent = false
+let numberOfArgsMethodTwo = -1
+let valueTwo
+let responseCorrelationIdTwo
+
+
+beforeAll( () => {
+
+    class MultiProvider {
+        multiMethodOne(...args) {
```
kpears201, line 40: where is the responder arg?
The Promise resolution triggers the ready call, not a response call

src/template/js/sdk/Capabilities/index.mjs
```diff
@@ -18,43 +18,82 @@
...
+
+            const providerCallArgs = []
+            
+            // only pass in parameters object if schema exists
+            if (imethod.parameters) {
+              providerCallArgs.push(request.parameters)
+            }
+
+            // only pass in the ready handshake if needed
+            if (imethod.handshake) {
```
kpears201, line 68: Did you swap this around where the ready callback is passed in and the promise resolution is the response?

src/template/js/sdk/Capabilities/index.mjs
```diff
@@ -18,43 +18,82 @@
...
+              })
+            }
+
+            const response = {
+              correlationId: request.correlationId
+            }
+
+            const result = provider[method].apply(provider, providerCallArgs).then(result => {
+              if (imethod.response) {
+                response.result = result
```
kpears201, line 82: did you change the name of response to result?

package.json
```diff
@@ -29,7 +27,8 @@
   "homepage": "https://github.com/rdkcentral/firebolt-core-sdk/blob/main/src/modules//ottx/firebolt-openrpc#readme",
   "devDependencies": {
     "jest": "^27.3.1",
-    "webpack-cli": "^4.8.0"
+    "webpack-cli": "^4.8.0",
+    "@firebolt-js/schemas": "../firebolt-schemas"
```
kpears201, line 31: why was this added? This would be a cyclical dependency since schemas depends on openrpc

src/template/js/sdk/Capabilities/index.mjs
```diff
@@ -18,43 +18,82 @@
...
+
+        const defined = !!methods.find(m => m === method) 
+
+        if (defined) {
+          Events.listen(module, `request${method.charAt(0).toUpperCase() + method.substr(1)}`, function (request) {
+
+            const providerCallArgs = []
+            
+            // only pass in parameters object if schema exists
+            if (imethod.parameters) {
```
kpears201, line 63: I don't think we want to change the number of args the provider method takes.
Seems like it should always take first arg as parameters. If no parameters are necessary this can just be null or empty object.

util/shared/modules.mjs
jlacivita: Doh, will fix.

util/shared/modules.mjs
jlacivita: you meant to put a `!` at the start of the `if`, right? Will do.

src/template/js/sdk/Capabilities/index.mjs
jlacivita: Roger

test/suite/provider.with-multiple-methods.test.js
jlacivita: I think we need to flip that.

Every Provider API needs to respond, but only some will have a ready method (those with the `x-handshake: true` tag extension)

Both of these methods have `x-handshake: false` (although leaving it off would work, too)

src/template/js/sdk/Capabilities/index.mjs
jlacivita: yup. see above thread.

src/template/js/sdk/Capabilities/index.mjs
jlacivita: Yes, to be consistent with the `polymorphic-pull` pattern.

We used request/response for the RPC objects that contain correlationId, etc.
We used parameters/result (borrowed from OpenRPC's conventions) for the actual things passed back and forth to the SDK.

package.json
jlacivita: Hmm, you're right. I used the shared `ProviderResponse` schema in my test module `provider.json`

Open to ideas here...

jlacivita: They're both "dev dependencies" so it's not actually a cyclical dependency for the artifacts we build, but I agree it points to some sort of issue worth figuring out.

src/template/js/sdk/Capabilities/index.mjs
jlacivita: yeah, I had it that way, and it felt weird to pass in `null`.

We're not really "changing" it. This code is just reusable code that all providers leverage. Once we have a release version of the schemas, we can't change it w/out a major version bump, so it felt cleaner to not add the noise of a `null` parameter that would never be useful.

kpears201: I meant it is odd that the author of the provider has to have knowledge of the API to create the method signature.
The signature could be:
```
my_method(parameters, handshake)
```
OR
```
my_method(handshake)
```
OR
```
my_method(parameters)
```
OR
```
my_method()
```
Depending on it's openrpc definition.
I feel a consistent API is better:
```
my_method(parameters, handshake)
```
Where parameters can just be null or empty object if it has no params.

jlacivita: Yeah, you've convinced me that parameters should always be there. So we'll have one signature:

```typescript
function my_method(parameters: object, handshake?: function)
```

parameters can be null if there are none, and handshake can be left off when it doesn't make sense.

src/template/js/sdk/Capabilities/index.mjs
kpears201: maybe it is fine, seems weird to use result when that is a term in JSONRPC. Quick glances, make it seem like a JSONRPC response instead of a request.

jlacivita: hmm, interesting point. Maybe we revisit when we go to 1.0? Too late to change `polymorphic-pull`

test/suite/provider.with-multiple-methods.test.js
kpears201: I don't understand why every provider doesn't just have a handshake.... doesn't hurt anything.

jlacivita: Exposing unneeded APIs can cause cause accidental usage. handshake is just for UX focus/visibility race conditions today. Any Providers that don't need that shouldn't get one for no reason. They might call it, and we'd ignore it for now.

But if we ever added another reason for a handshake, they'd be calling it already, and hard to track down side effects might happen.

jlacivita: Also, each capability will have it's own provider documentation, so this stuff will be front and center once we have docs. No point in having it there all the time and saying "Ignore this" in half of our Providers.

kpears201: I see it as the opposite, the pattern is exactly the same for ever provider.
```
async my_method(parameters, responder) {
  responder(response)
}
```
Since it is an async method, it is all taken care of for them.
If it doesn't require a handshake, then they just don't do anything, they still just call responder with the response.
If it does require a handshake, they do the exact same thing, just need to `await` in this method until their UI is ready.

I see your implementation as the opposite, they now have to consciously call the handshake callback before they resolve the promise.

jlacivita: I think it's easier to explain why I switched it with more realistic examples.

If we're doing a provider method for the RPC `Device.onRequestName`, then it's more natural to _return_ the actual name, since that's how methods work in virtually all programming languages, JS included. Also note that a method like this wouldn't have any parameters:

```javascript
function name() {
  return Promise.resolve(name)
}
```

This looks much like any getter in any system would look like.

The way you're proposing, they'd have to write it like this:

```javascript
function name(parameters, responder) {
  responder(name)
}
```

Doing it the above way requires:

1. adding a parameter to their method that is useless (`parameters`)
2. handling another parameter just to return a value in a strange way that we made up

Now if it's a more UX-centric method, like `PinChallenge.onRequestChallenge` then they're going to need a handshake, but any UX is almost certain to have parameters. Again, having the return value be the result and the handshake in the params:

```javascript
async function challenge(parameters, handshake) {
  let pinPromise = await showSomeUX() // returns a promise chain: 1st for UX ready, 2nd for pin string
  handshake()
  let pin = await pinPromise
  return pin
}
```

If we flip it around and pass the responder:

```javascript
async function challenge(parameters, responder) {
  let pinPromise = await showSomeUX() // returns a promise chain: 1st for UX ready, 2nd for pin string
  handshake()
  // can't do await here, or UX won't show up...
  pinPromise.then( pin => {
    responder(pin)
  })
  return pinPromise
}
```

In the first example, the app partner only has to care about asynchronous details that they choose to implement plus the fact that ALL Firebolt APIs return promises.

In the second, they have to:

1. use only `.then()` syntax (or mix `await` / `then()`)
2. Return "something." Our spec says nothing, but at this point, they'll possibly have their own `Promise` that likely resolves to what we want.
3. Send the result to us with the `responder` (inside a `then()`) at the same time they just returned for the handshake

Yes, it's more consistent from provider to provider, but it's also more complicated.

I'm suggesting that "provider" is a pattern, not an API, so consistency isn't the top concern.

If it's not consistently, it should be simplicity, so I think we should have separate APIs that fit into the "provider" pattern, with details that match that specific API.

jlacivita: @kpears201 & I met offline and discussed further.

Summary:

- Agree that provider APIs should return a Promise that resolves to the provided "thing"
- Need to account for a UX "ready" handshake in a forward thinking way that doesn't expose new APIs to apps that aren't fulfilling provider APIs

So the new JavaScript signature will be:

```javascript
async function challenge(parameters, session) {
  let pinPromise = await showSomeUX()
  session.focus()
  return pinPromise
}
```

The `session` parameter looks like:

```typescript
interface ProviderSession {
  correlationId(): string // return the current correlationId for debugging
  focus?(): void // requests focus for the current provider
}
```

The `focus()` method will only be available on providers that have the x-allow-focus tag extension.

This is a good place for extending the API, e.g. we could add `session.requestor()` in the future to expose which app initiated the provider flow.

jlacivita: Also, really simple providers can still be very clean:

```javascript
function name() {
  return Promise.resolve('Living Room')
}
```

jlacivita: Resolving. Reopen if there's more to discuss

src/js/shared/Capabilities/index.mjs
```diff
@@ -18,44 +18,97 @@
...
+
+  if (provider.constructor.name !== 'Object') {
+    methods.push(...Object.getOwnPropertyNames(Object.getPrototypeOf(provider)).filter(item => typeof provider[item] === 'function' && item !== 'constructor'))
+  }
+  else {
+    methods.push(...Object.getOwnPropertyNames(provider).filter(item => typeof provider[item] === 'function'))
+  }
+
+  // make sure every interfaced method exists in the providers methods list
+  const valid = iface.every(method => methods.find(m => m === method.name.split('.').pop()))
```
kpears201, line 45: This should go below the iface existing check

jlacivita: Good catch, fixed

kpears201 (approved):

PR #35, opened 2022-05-20
TemporalSet Method Templates

jlacivita:
This feature adds a new method tag/template called `temporal-set`.

A Temporal Set is an Array-like data structure that only contains unique items, i.e., a "set," and has it's contents change over an indefinite length of time, i.e. "temporal."

This supports generated JavaScript APIs like this:

```javascript
const process = Module.find("Firebolt Tool",
  result => {
    console.log('this item shows up')
  },
  result => {
    console.log('this item goes away')
  }
)

setTimeout( () => { process.stop() }, 10000)

```

From a single human-authored OpenRPC method (with the supporting RPC methods/events auto-generated)

### DIFF
--- a/src/js/shared/Transport/MockTransport.mjs
+++ b/src/js/shared/Transport/MockTransport.mjs
@@ -34,6 +34,13 @@ if (win.__firebolt && win.__firebolt.testHarness) {
 
 function send(message) {
   let json = JSON.parse(message)
+
+  // handle bulk sends
+  if (Array.isArray(json)) {
+    json.forEach(j => send(JSON.stringify(j)))
+    return
+  }
+
   let [module, method] = json.method.split('.')
 
   if (testHarness && testHarness.onSend) {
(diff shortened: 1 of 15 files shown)


src/js/shared/Transport/MockTransport.mjs
```diff
@@ -34,6 +34,13 @@ if (win.__firebolt && win.__firebolt.testHarness) {
 
 function send(message) {
   let json = JSON.parse(message)
```
michaelcfine, line 36: Technically not 'json'... an actual object. E.g., "obj".

src/schemas/firebolt-openrpc.json
```diff
@@ -302,6 +314,29 @@
...
+            "type": "object",
+            "properties": {
+                "tags": {
+                    "allOf": [
+                        {
+                            "type": "array",
+                            "minItems": 1,
+                            "maxItems": 1,
+                            "items": {
+                                "type": "object"
```
michaelcfine, line 327: Are we saying to use temporal sets, you must use an object (not, say, just strings)? Probably fine, but we should document somehow.

michaelcfine (approved):
👍 Looks good, with just two minor comments.

PR #40, opened 2022-06-09
Map integer type to Number instead of bigint

jlacivita:
bigint isn't valid JSON, so all of these schemas with bigint were actually impossible to use


jlacivita:
Certification Suite passed, merging

PR #41, opened 2022-06-13
Wire up the correlationId to pull APIs

jlacivita:


kpears201:
This is not backwards compatible and will break XClass

kpears201 (approved):

jlacivita:
Worked with @kpears201 to name the variables in a way that only breaks unimplemented use cases.

PR #43, opened 2022-06-15
fix: Add jsonrpc as a protocol for websocket transport

kpears201:


jlacivita (approved):
Looks good to me. I tried to find a list of standard values, but could not... oh well.

PR #44, opened 2022-06-15
fix: Property setter jsonrpc should follow set{Prop} pattern

kpears201:


src/js/shared/Transport/MockTransport.mjs
```diff
@@ -33,6 +33,7 @@ if (win.__firebolt && win.__firebolt.testHarness) {
 }
 
 function send(message) {
+  console.debug('Sending message to transport: ' + message)
```
jlacivita, line 36: Did you mean to leave this here?

jlacivita (approved):

src/js/shared/Transport/MockTransport.mjs
jlacivita: It's in MockTransport, so we can discuss later. merging to unblock team

jlacivita (approved):

PR #45, opened 2022-06-29
fix: Open-rpc param name should be value for setters

kpears201:


kpears201:
> Why not just use the name attribute from the getter's result?

That is what is there now and is wrong. Our docs and the SDK have it send the value as "value" not the name attribute. This is to fix the openrpc file to match the docs and sdk

jlacivita:
ah ok. works for me

jlacivita (approved):

PR #46, opened 2022-07-08
fix: move any code execution outside of global execution path. Lazily…

kpears201:
… init transport


jlacivita (approved):
Approving. we def need QA on this, though!

PR #47, opened 2022-07-11
Request Methods

jlacivita:
A Response Method is an RPC method that is attached to an RPC result by the SDK for convenience. This allows for results that represent their own focused APIs. For example:

```javascript
const pet = House.getPet()
pet.walk()
```

In the example above, pet.walk() is an RPC method that takes a pet object as a parameter. Calling pet.walk() via the Response Method syntax allows contextual APIs to be called with simpler method signatures. These methods may or may not be hidden via the rpc-only tag to avoid Apps calling them without valid parameters, if desired.


kpears201:
home come transforms has to happen within the Transport module?
```
const transforms = {"methods":{"aMethod":{"type":"string","x-method":"Simple.method","x-this-param":"accessory"}}}
return Transport.send('simple', 'methodWithMethodAttribute', { title }, transforms)
```
 vs.
```
const transforms = {"methods":{"aMethod":{"type":"string","x-method":"Simple.method","x-this-param":"accessory"}}}
return Transport.send('simple', 'methodWithMethodAttribute', { title }).then(x => Results.transform(x, transforms))
```

Then you don't have to pass transforms through a bunch of methods and it stays localized to the thing that needs the transforming , the module

PR #52, opened 2022-08-15
Extensions Module

jlacivita:
This branch is locked, and features where cherry-picked into the `feature/http-methods` branch.


src/js/shared/Extensions/index.mjs
```diff
@@ -0,0 +1,50 @@
...
+
+    if (queue[id]) {
+        initialize(id, queue[id])
+        delete queue[id]
+    }
+}
+
+// Method for handing off platform tokens to extension SDKs
+registerAPI('token', _ => Transport.send('authentication', 'token', { type: 'platform' }))
+registerAPI('authorize', grants => Transport.send('capabilities', 'request', {grants} ))
```
kpears201, line 22: I don't understand why these are two different methods? Doesn't token initiate the grants and return a token?

jlacivita: I waffled on this as well.

Token just returns a token. It has no way to know which grants are needed. It's wired up to `Authentication.token('platform')`.

For extension SDKs, we basically need the SDK to tell us which capabilities any REST APIs will use, so there needs to be a way to list them out in a method that triggers SUGs.

We could absolutely make a token method that _also_ takes a list of role/caps that are needed, but I didn't want to add more params to `Authentication.token()`, but having it as a separate method allows for other use cases, e.g. Apps that want to request all of the SUGs up front, instead of waiting for the APIs to be hit.

PR #53, opened 2022-08-15
Http Methods

jlacivita:
Creates the ability to add HTTP/REST methods to an SDK. Would be used by Extension SDKs

Creates a new module called `Http` that looks like a Transport Layer.

Methods that have the "http" tag will:

```
import Transport from '../Http'
```

So that existing method templates will just work over Http.

The "http" tag supports several extensions:

- x-http-endpoint - The baseUrl for the REST API
- x-http-path - the path to append to the baseUrl
- x-http-headers - any headers to include
- x-http-parameters - any query parameters to include

Additionally, all of these extensions will support the following macros:

- ${param.<name>} - insert the value of param <name>
- ${token} - insert the value of the Extension SDK's token


test/Server.js
```diff
@@ -0,0 +1,89 @@
...
+            response.setHeader("Content-Type", "application/json")
+            response.writeHead(200)
+
+            if (method) {
+                response.write(JSON.stringify(method.examples[0].result.value))
+            }
+            else if (request.url.endsWith('/account/authenticate')) {
+                response.write(JSON.stringify({
+                    oat: "OAT",
+                    bearerToken: "BEARER",
```
mhughesacn, line 67: Hi Jeremy: Please confirm that these two lines are not sensitive.  (Unlikely but I have to ask.)
Thanks

src/js/shared/Http/index.mjs
```diff
@@ -0,0 +1,123 @@
...
+function send(module, method, params, http = {}) {
+
+    let path = http.path || `/${module}/${method}`
+    const headers = http.headers || {}
+    const body = params && JSON.stringify(params) || null
+    let query = http.parameters || ''
+    
+    Object.entries(params).forEach(([name, value]) => {
+        const find = '${param.' + name + '}'
+        path = path.replace(find, value)
```
kpears201, line 54: This will only replace a single occurrence, guess that is probably the usual case, but might be better to do a replace all?
However, I found this out the hard way, WPE does not support replaceAll function.
You have to do this:
```
const re = new RegExp('\\${param.' + name + '}', 'g')
path = path.replace(re, value)
```

src/js/shared/Http/index.mjs
```diff
@@ -0,0 +1,123 @@
...
+        query = query.replace(find, value)
+        Object.keys(headers).forEach(header => {
+            headers[header] = headers[header].replace(find, value)
+        })
+    })
+
+    return new Promise((resolve, reject) => {
```
kpears201, line 61: might be better to rewrite this as an async function, a lot of nested `.then`s going on

util/sdk/macros/index.mjs
```diff
@@ -185,13 +214,31 @@ const makeEventName = x => x.name[2].toLowerCase() + x.name.substr(3) // onFooBa
...
-const generateAggregateMacros = (modules = {}) => Object.values(modules)
-  .reduce((acc, module) => {
-    acc.exports += `export { default as ${getModuleName(module)} } from './${getModuleName(module)}/index.mjs'\n`
-    acc.mockImports += `import { default as ${getModuleName(module).toLowerCase()} } from './${getModuleName(module)}/defaults.mjs'\n`
-    acc.mockObjects += `  ${getModuleName(module).toLowerCase()}: ${getModuleName(module).toLowerCase()},\n`
+const generateAggregateMacros = (modules = {}, packageJson) => Object.values(modules)
+  .reduce((acc, module, i, arr) => {
+    acc.imports += `${generateAggregateImports(module)}\n`
+
+//    import _Commerce from './Commerce/index.mjs'
```
kpears201, line 221: should be removed?

util/sdk/macros/index.mjs
```diff
@@ -354,8 +416,38 @@ const generateImports = json => {
...
 
-const generateInitialization = json => generateEventInitialization(json) + '\n' + generateProviderInitialization(json) + '\n' + generateDeprecatedInitialization(json)
+const generateInitialization = json => [generateEventInitialization(json), generateProviderInitialization(json), generateDeprecatedInitialization(json)].join('\n')
+
+const generateConfig = (json, pkg) => [generateHttpConfig(json, pkg)].join('\n')
+
+const generateHttpConfig = (json, pkg) => {
+  if (httpMethodsOrEmptyArray(json).length) {
+    const config =
+`Configuration.set('apiBaseUri', '${json.info["x-http-endpoint"]}')
```
kpears201, line 426: This is teetering on enough code to be in a js file instead of a string

src/js/shared/Http/index.mjs
```diff
@@ -0,0 +1,123 @@
...
+                options.body = body
+            }
+
+            proxyProvider(resource, options).then( ({ resource, options }) => {
+                console.log(`Fetch ${method} ${resource}`)
+                options.body && console.log(options.body)
+
+                fetch(resource, options).then(response => {
+                    if (response.ok) {
+                        response.json().then(json => {
```
kpears201, line 97: So extension API endpoints must be json? Guess that is fine

util/sdk/macros/index.mjs
```diff
@@ -354,8 +416,38 @@ const generateImports = json => {
...
+
+function _initialize(config, apis) {
+  Configuration.set(config)
+  Http.setEndpoint(Configuration.get('apiBaseUri'))
+  Http.onToken(apis.token)
+  Http.onAuthorize(apis.authorize)
+}
+
+export const initialize = window.__firebolt ? () => { throw new Error('Use Extensions.initialize() from \\'@firebolt-js/sdk\\' to initialize ${pkg.name}.') } : _initialize
```
kpears201, line 435: think I might need this walked through. Why would the initialize method throw an error if __firebolt is not undefined?

test/Server.js
jlacivita: Correct, they are not. The code deals with tokens, but there are no secrets or example values present.

src/js/shared/Http/index.mjs
jlacivita: Thanks for the heads up. we should file a bug with WPE, too.

src/js/shared/Http/index.mjs
jlacivita: Stylistically, I agree.

However, I've been resisting the urge to use `async/await` because we haven't used them yet in the code. I just did a quick search and saw it's used in the new version of Lifecycle.ready() in the Core SDK. Have we tested that code on all of the browsers we care about?

util/sdk/macros/index.mjs
jlacivita: Yup, thanks

util/sdk/macros/index.mjs
jlacivita: Absolutely agree. I think my next branch will be to remove every bit of hard coded JavaScript from our codebase, which is the first step towards enabling other language SDKs to be generated, e.g. RUST or C.

I added a `src/template/js/configuraiton` directory that works the same way as `src/templates/js/methods` and put this code in it under `http.js` (with an empty `default.js`).

src/js/shared/Http/index.mjs
jlacivita: I think so? We leverage JSON-Schema so heavily in both the SDK, Docs, and Certification, that I think we should stick with it unless there's good reason.

util/sdk/macros/index.mjs
jlacivita: This `export const initialize` is for non-Firebolt apps that need to use the SDK. They can import the initialize method and use the SDK w/out Firebolt. Not a high priority use case, but seemed like a good bonus to get teams to leverage our framework.

Firebolt get's access to the _real_ initialize method (not the exported one) via `window.__firebolt.registerExtensionSDK`.

PR #54, opened 2022-08-17
fix: docs for jsonrpc in provider

kpears201:


kpears201:
The docs were mismatched from the impl: https://github.com/rdkcentral/firebolt-openrpc/blob/main/src/template/js/sdk/Capabilities/index.mjs#L110

jlacivita (approved):

PR #56, opened 2022-08-22
fix: multiple provider doc headers for same module are broken

kpears201:


kpears201:
Most important is the newline at the end of the template. Without it, modules that have multiple provider capabilities were causing the header to not be rendered as a header.
I also fixed the ** around JSON-RPC. Looks like that was just done for provider template. Not sure if that was intended to make the text bold, but it didn't work since this was html. We can make the text bold using html, but none of the other templates has JSON-RPC bold.

jlacivita (approved):

PR #58, opened 2022-08-25
Context Parameters

jlacivita:


src/js/shared/Events/index.mjs
```diff
@@ -64,7 +72,9 @@ const doListen = function(module, event, callback, once) {
 
     let events = (event === '*' ? validEvents[module] : [event]) // explodes wildcards into an array
     let promises = []
-    const key = module + '.' + event // this might be a wildcard, e.g. 'lifecycle.*'
+    const hasContext = Object.values(context).length > 0
+    const contextKey = Object.keys(context).sort().map(key => key + '=' + JSON.stringify(context[key])).join('&')
```
kpears201, line 76: I don't quite understand how this works.
The callback is stored now using the contextKey, but when emitting the event, I don't think it uses this same key to find the callback?

src/js/shared/Prop/index.mjs
```diff
@@ -1,17 +1,19 @@
...
-  if (args.length === 0) {
+function prop(moduleName, key, params, callbackOrValue, immutable, readonly, contextParameterCount) {
+  const numArgs = Object.values(params).length
+
+  if (numArgs === contextParameterCount && !callbackOrValue) {
     // getter
-    return Transport.send(moduleName, key)
-  } else if (args.length === 1 && typeof args[0] === 'function') {
+    return Transport.send(moduleName, key, )
+  } else if (numArgs === (contextParameterCount) && typeof callbackOrValue === 'function') {
```
kpears201, line 10: these parenthesis are unnecessary here

src/js/shared/Events/index.mjs
jlacivita: You might be right. let me look closer at my tests / code.

src/js/shared/Prop/index.mjs
jlacivita: will fix.

src/js/shared/Events/index.mjs
jlacivita: Yeah, it was totally broken. I had to refactor the Event/Transport integration, but it's much more solid now

PR #59, opened 2022-09-16
Added fix for FIRESDK-20 and FIRESDK-21

ChakrabortySubhrajyoti:


util/docs/macros/index.mjs
```diff
@@ -1006,6 +1006,12 @@ function generateJavaScriptExample(example, m, moduleJson = {}, templates = {})
     const template = m.tags && m.tags.map(t=>t.name).find(t => Object.keys(templates).includes('examples/' + t + '.md')) || 'default'
     typescript = templates[`examples/${template}.md`]
 
+    // added null check to handle the scenario when parameter is optional
```
jlacivita, line 1009: The original code is sort of badly designed, and this fix makes it a bit worse.

We should probably refactor `formatParams` to leave out params that are explicitly in the example, and allow for examples to explicitly set a param value to `null` if it wants to.

jlacivita (changes requested):

ChakrabortySubhrajyoti:
Closing this PR, as the changes for FIRESDK-20 in WIP.

PR #63, opened 2022-09-27
Worked on generating proper link

ChakrabortySubhrajyoti:
added asPath and baseUrl to generate proper URL.

### DIFF
--- a/util/shared/typescript.mjs
+++ b/util/shared/typescript.mjs
@@ -387,7 +387,7 @@ function getSchemaShape(moduleJson = {}, json = {}, schemas = {}, name = '', opt
       else {
         // grab the type for the non-array schema, so we get the link for free
         // added asPath and baseUrl to generate proper URL
-        let type = getSchemaType(module, json.items, schemas, {code: options.code, link: options.link, title: options.title, asPath: options.asPath, baseUrl: options.baseUrl})
+        let type = getSchemaType(module, json.items, schemas, options)
         // insert the [] into the type
         if (options.link) {
           type = type.replace(/\[(`?)(.*)(`?)\]/, '\[$1$2\[\]$3\]')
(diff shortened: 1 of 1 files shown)


util/shared/typescript.mjs
```diff
@@ -386,7 +386,8 @@ function getSchemaShape(moduleJson = {}, json = {}, schemas = {}, name = '', opt
       }
       else {
         // grab the type for the non-array schema, so we get the link for free
-        let type = getSchemaType(module, json.items, schemas, {code: options.code, link: options.link, title: options.title})
+        // added asPath and baseUrl to generate proper URL
+        let type = getSchemaType(module, json.items, schemas, {code: options.code, link: options.link, title: options.title, asPath: options.asPath, baseUrl: options.baseUrl})
```
cletustboone, line 390: Is there any reason this can't be:
```
let type = getSchemaType(module, json.items, schemas, options)
```

util/shared/typescript.mjs
```diff
@@ -386,7 +386,8 @@ function getSchemaShape(moduleJson = {}, json = {}, schemas = {}, name = '', opt
       }
       else {
         // grab the type for the non-array schema, so we get the link for free
-        let type = getSchemaType(module, json.items, schemas, {code: options.code, link: options.link, title: options.title})
+        // added asPath and baseUrl to generate proper URL
```
cletustboone, line 389: What does this have to do with `asPath` and `baseUrl`?

cletustboone: I think you're saying those get passed in and used by `getSchemaType` because they are part of the `options` hash, right?

cletustboone (approved):

util/shared/typescript.mjs
ChakrabortySubhrajyoti: Yes .

PR #68, opened 2022-10-31
fix: Interfaceless providers

jlacivita:
Allows an API to provide a capability with no provider interface, e.g. for push/pull APIs.

### DIFF
--- a/src/js/shared/Events/index.mjs
+++ b/src/js/shared/Events/index.mjs
@@ -229,7 +229,6 @@ export const prioritize = function(...args) {
 }
 
 const unsubscribe = (key) => {
-  console.log(key)
   const [module, event] = key.split('.').slice(0, 2)
   Transport.send(module, 'on' + event[0].toUpperCase() + event.substr(1), { listen: false })
 }
--- a/util/shared/modules.mjs
+++ b/util/shared/modules.mjs
@@ -54,26 +54,33 @@ const getMethods = compose(
     getPath(['methods'])
 )
 
-const isProviderMethod = compose(
-    option(false),
-    map(_ => true),
-    chain(
-      find(
-        and(
-          propEq('name', 'capabilities'),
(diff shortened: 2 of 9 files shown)


util/shared/modules.mjs
```diff
@@ -54,26 +54,33 @@ const getMethods = compose(
...
 
-const isProviderMethod = compose(
-    option(false),
-    map(_ => true),
-    chain(
-      find(
-        and(
-          propEq('name', 'capabilities'),
-          propSatisfies('x-provides', not(isEmpty))
+const isProviderInterfaceMethod = compose(
```
cletustboone, line 57: 💪🏻

cletustboone:
I'll ask the dumb question. Why are push/pull APIs interfaceless?

cletustboone (approved):

src/js/shared/Events/index.mjs
```diff
@@ -229,7 +229,6 @@ export const prioritize = function(...args) {
 }
 
 const unsubscribe = (key) => {
-  console.log(key)
```
cletustboone, line 232: Do you really want this `console.log` here?

jlacivita:
Techincally, they're "Provider-interfaceless" not "interfaceless"

A provider interface is an interface that the app implements with their own custom code, so we can call APIs on it. Effectively a reverse interface, made up of a bunch of private events that trigger calling the APIs, and provide response/error methods for each Provider Interface API.

A push/pull API is just a normal interface that is part of the RPC interface described by our OpenRPC docs.

src/js/shared/Events/index.mjs
jlacivita: no, that's why I took it out.

cletustboone: Oh my bad...jeez, you took it out. I read it as an addition.

PR #71, opened 2022-11-03
Added support for null params

ChakrabortySubhrajyoti:


cletustboone:
Just, FYI, I'm waiting until we can merge #74 because it contains a degree of sanity check to make sure these changes haven't changed the behavior of the generated SDK.

PR #88, opened 2023-04-19
Fix/2.0 bugs next

jlacivita:
- Fixed context parameters on events and properties
- Reintroduced per-module Events enum
- Emit [@deprecated](https://github.com/deprecated) as appropriate
- renamed a few static params, e.g. subscriber vs callback
- Fixed type of Module.listen params
- Fixed AudioProfiles object property names
- Fixed subscriber callback param name
- Fixed various ProgramEntity types programType field to be appropriate cons value


kpears201 (dismissed):

kevinshahfws (dismissed):

languages/javascript/templates/modules/index.mjs
```diff
@@ -24,7 +24,6 @@
 
 export default {
 
-  /* ${EVENTS_ENUM} */
```
kevinshahfws, line 27: Please put this line back

jlacivita: done

PR #90, opened 2023-05-04
fix: Sort the macrofied schemas based on schema refs

sramani-metro:


jlacivita (changes requested):

src/macrofier/engine.mjs
```diff
@@ -625,21 +625,36 @@ function generateSchemas(json, templates, options) {
...
 
+  const list = []
+
   // schemas may be 1 or 2 levels deeps
   Object.entries(schemas).forEach( ([name, schema]) => {
     if (isSchema(schema)) {
-      generate(name, schema)
+      list.push([name, schema])
     }
     else if (typeof schema === 'object') {
```
jlacivita, line 635: Not directly related to your change, but I think we can get rid of this entire `else` block. It's for external schemas, that won't get output by the current module. Your code below doesn't account for these, but we actually filter them out immediately up at the top of the code after calling this method

Probably makes sense to leave your sorter method the way it is, and remove this else block.

src/macrofier/engine.mjs
```diff
@@ -625,21 +625,36 @@ function generateSchemas(json, templates, options) {
...
         if (name !== 'uri') {
-          generate(name, schema, uri)
+          list.push([name, schema, uri])
         }
       })
     }
   })
 
+  list.sort((a, b) => {
+    const aInB = isDefinitionReferencedBySchema('#/components/schemas/' + a[0], b[1])
```
jlacivita, line 646: You'll need to check for the presence of `uri` (third array item).

If it's present, then the path to the schema is `/x-schemas/${module.info.title}` and we didn't grab the title in the loop above, so we don't have it down here.

I recommend removing the else above, to keep this code simpler.

jlacivita (changes requested):
Sorry, one more minor change, to avoid confusion/bugs in the future. See comments.

jlacivita (approved):
Looks good, thanks

PR #93, opened 2023-05-15
Enum fixes

HaseenaSainul:
1. Description added
2. Removed last ',' enum list
3. Template alignment to get proper indentation in generation 
4. Support added to generate enum inside properties as well


src/macrofier/engine.mjs
```diff
@@ -402,6 +406,7 @@ const insertMacros = (fContents = '', macros = {}) => {
   fContents = fContents.replace(/[ \t]*\/\* \$\{TYPES\} \*\/[ \t]*\n/, macros.types)
   fContents = fContents.replace(/[ \t]*\/\* \$\{PROVIDERS\} \*\/[ \t]*\n/, macros.providerInterfaces)
   fContents = fContents.replace(/[ \t]*\/\* \$\{ENUMS\} \*\/[ \t]*\n/, macros.enums)
+  fContents = fContents.replace(/[ \t]*\/\* \$\{ENUMS_CONVERSION\} \*\/[ \t]*\n/, macros.enumsConversion)
```
jlacivita, line 409: So this purpose of this macro is to generate code that enables converting an `enum` from C into the JSON library that we are leveraging in C?

Is there a way to generate this conversion based on the `destination` context? We know which template file we are inserting the schemas into based on `options.destination`, so if we can detect it, instead of creating another macro, that would be ideal.

HaseenaSainul: This is to generate enum value string as like below,
ENUM_CONVERSION_BEGIN(SecondScreen_type)
        { SECONDSCREEN_TYPE_DIAL, _T("dial") }
ENUM_CONVERSION_END(SecondScreen_type)

This is cpp specific part and it has to be under WPEFramework namespace, since it is generated based on thunder json module format. https://github.com/rdkcentral/firebolt-openrpc/pull/93/files#diff-97dd97692f7573b953d3489feab237c014a7b0719d301ce1fbd1ec15f9f2e368R33

Hence added a specific macro ENUMS_MACRO

jlacivita: So the Module.cpp template does not have any instance of the existing ${ENUMS} macro.

Therefor, we can use the `${ENUMS}` macro and check the value of `options.destination`. If it's `Module.cpp` then we insert the conversion with the code in this PR. 

That way we do not introduce a new macro just to fulfill a particular implementation pattern that we've chosen.

Can we drop the `${ENUMS_CONVERSION}` macro and add logic to do an `if/else` on `options.destination`?

HaseenaSainul: yes, I have removed ENUMS_CONVERSION and used ENUMS itself based on destination values

jlacivita (changes requested):

src/macrofier/engine.mjs
```diff
@@ -486,8 +488,66 @@ function insertTableofContents(content) {
...
+  const template = getTemplate(templateName, templates).split('\n')
+  let schema = isEnumType(sch) ? sch : getEnumProperties(sch)
+  for (var i = 0; i < template.length; i++) {
+    if (template[i].indexOf('${key}') >= 0) {
+      template[i] = schema.enum.map(value => {
+        const safeName = value.split(':').pop().replace(/[\.\-]/g, '_').replace(/\+/g, '_plus').replace(/([a-z])([A-Z0-9])/g, '$1_$2').toUpperCase()
+        return template[i].replace(/\$\{key\}/g, safeName)
+                          .replace(/\$\{value\}/g, value)
+      }).join('\n')
+      if (!templateName.includes("enum_conversion")) {
```
jlacivita, line 537: What's this code doing?

src/macrofier/engine.mjs
```diff
@@ -496,19 +557,18 @@ const generateEnums = (json, templates) => {
   return compose(
```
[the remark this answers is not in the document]
HaseenaSainul: sure

languages/c/src/types/JSONHelpers.mjs
```diff
@@ -42,15 +42,15 @@ ENUM_CONVERSION_END(Advertising_SkipRestriction)
 // TODO - this should be a global function in the main /src/ directory... not part of a language pack
 const keyName = val => val.split(':').pop().replace(/[\.\-]/g, '_').replace(/\+/g, '_plus').replace(/([a-z])([A-Z0-9])/g, '$1_$2').toUpperCase()
 
-function getJsonEnumConversion(schema, module, { name }) {
-    name = capitalize(schema.title || name)
-    let e = `ENUM_CONVERSION_BEGIN(${module.info.title}_${name})\n`
-
+function getJsonEnumConversion(schema, { name }) {
```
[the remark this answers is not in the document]
HaseenaSainul: it is called line @327 based on destination values, if add third param, it is not required then. Will update based on that

src/macrofier/engine.mjs
HaseenaSainul: excluding the last line comma replacement for enumconversion case, since it need last comma also

src/macrofier/engine.mjs
```diff
@@ -324,8 +324,7 @@ const generateMacros = (obj, templates, languages, options = {}) => {
 
   const imports = generateImports(obj, templates)
   const initialization = generateInitialization(obj, templates)
-  const enums = generateEnums(obj, templates)
-  const enumsConversion = generateEnumsConversion(obj, templates)
+  const enums = (options.destination && options.destination.endsWith(".cpp") ? generateEnumsConversion(obj, templates) : generateEnums(obj, templates))
```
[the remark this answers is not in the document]
HaseenaSainul: updated to pick based on suffix.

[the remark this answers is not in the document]
HaseenaSainul: But the template replacement part is inside convertEnumTemplate, since it is handling enums from array, properties etc too. But I have taken care the main points you have suggested here

src/macrofier/engine.mjs
```diff
@@ -486,29 +488,77 @@ function insertTableofContents(content) {
...
   map(([_, val]) => val),
   filter(([_key, val]) => isObject(val))
 )
 
-const generateEnums = (json, templates) => {
+const generateEnums = (json, templates, options = { destination: '' }) => {
+  const suffix = options.destination.split('.').pop()
   return compose(
     option(''),
+    map(val => val ? (suffix && getTemplate(`/sections/enum.${suffix}`, templates) ? getTemplate(`/sections/enum.${suffix}`, templates).replace(/\$\{schema.list\}/g, val.trimEnd()) : val) : ''),
```
jlacivita, line 559: Having trouble understanding this line. Are you trying to dynamically swap which section template you use based on suffix? If so, you'll need to support cases where you don't know the suffix (i.e. just calling `getTemplate('/sections/enum')`.

It's probably easier to just always grab the suffix-less template for sections. We can add suffix-aware templates as needed to avoid the code getting overly complex.

jlacivita: Oh, and there's no template in your PR for `/sections/enum.*` I think you forgot to add new files.

src/macrofier/engine.mjs
```diff
@@ -486,29 +488,77 @@ function insertTableofContents(content) {
...
-          template[i] = schema.enum.map(value => {
-            const safeName = value.split(':').pop().replace(/[\.\-]/g, '_').replace(/\+/g, '_plus').replace(/([a-z])([A-Z0-9])/g, '$1_$2').toUpperCase()
-            return template[i].replace(/\$\{key\}/g, safeName)
-                              .replace(/\$\{value\}/g, value)
-          }).join('\n')
-        }
-      }
-      return template.join('\n').replace(/\$\{name\}/g, schema.title).replace(/\$\{NAME\}/g, schema.title.toUpperCase())
-    })),
+    map(map((schema) => convertEnumTemplate(schema, suffix ? `/types/enum.${suffix}` : '/types/enum', templates))),
```
jlacivita, line 561: This looks good, but the new files `/types/enum.cpp` and `/types/enum.h` are missing in the PR

jlacivita (changes requested):
We're super close.

Minor issue in the enums section template selection, and a couple template files that I suspect you forgot to add to the PR.

languages/c/src/types/JSONHelpers.mjs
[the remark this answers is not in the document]
jlacivita: Sorry but I'm still not seeing where this method `getJsonEnumConversion` is imported and called.

Which file imports it?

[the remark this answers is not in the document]
HaseenaSainul: It is not used anywhere now, so removed

src/macrofier/engine.mjs
jlacivita: Got it... let's get this PR merged. at some point I'd like to add a ${delimiter}${end.delimiter} macro, that we can wrap around the comma in the template, and then just remove any text between those macros for the last item.

No need to do it now.

jlacivita (approved):
All good, thanks!

HaseenaSainul:
Thanks @jlacivita

PR #94, opened 2023-05-18
Types and Accessor related fixes

HaseenaSainul:
1. Accessor creation
2. JsonContainer creation
3. Types order placed based on depenedencies
4. filtered redundant new lines
5. UnamedSchema issue fixes


src/macrofier/engine.mjs
```diff
@@ -576,6 +635,7 @@ function generateSchemas(json, templates, options) {
     }    
 
     let content = getTemplate('/schemas/default', templates)
+    content = content.trim()
```
jlacivita, line 638: It can cause undesired side-effects to trim whitespace from the content. For example Markdown is sometimes picky about how much whitespace is needed for a document to render properly.

Can we remove this?

src/macrofier/engine.mjs
```diff
@@ -1003,9 +1061,9 @@ function insertMethodMacros(template, methodObj, json, templates, examples={}) {
     .replace(/\$\{method\.result\.name\}/g, result.name)
     .replace(/\$\{method\.result\.summary\}/g, result.summary)
     .replace(/\$\{method\.result\.link\}/g, getLinkForSchema(result, json)) //, baseUrl: options.baseUrl
-    .replace(/\$\{method\.result\.type\}/g, types.getSchemaType(result.schema, json, {title: true, asPath: false, destination: state.destination })) //, baseUrl: options.baseUrl
-    .replace(/\$\{event\.result\.type\}/, isEventMethod(methodObj) ? types.getSchemaType(result.schema, json, { destination: state.destination, event: true, description: methodObj.result.summary, asPath: false }): '') //, baseUrl: options.baseUrl
-    .replace(/\$\{method\.result\}/g,  generateResult(result.schema, json, templates))
```
jlacivita, line 1008: was this not working before?

src/macrofier/engine.mjs
```diff
@@ -1133,45 +1191,49 @@ function insertExampleMacros(template, examples, method, json, templates) {
   return template.replace(/\$\{method\.examples\}/g, content)
 }
 
-function generateResult(result, json, templates) {
-  const type = types.getSchemaType(result, json, { destination: state.destination, section: state.section  })
+function generateResult(schema, json, templates) {
+
+  let name = schema.name || schema.title
```
jlacivita, line 1196: Hmm... schemas never have a `name` attribute, only `title`. Are we adding the name to it somewhere else?

Ideally we'd pass it in as a parameter to this method, instead of mutating the schema object (which could cause side effects elsewhere in the code).

src/macrofier/engine.mjs
```diff
@@ -1133,45 +1191,49 @@ function insertExampleMacros(template, examples, method, json, templates) {
...
+  let result = schema.schema ? schema.schema : schema
```
jlacivita, line 1197: We shouldn't add this code to any methods. I know I left some inspiration to do this in other places, but it makes the code harder to maintain if the `result` parameter could either be a JSON-Schema schema or an OpenRPC Content Descriptor object (https://spec.open-rpc.org/#content-descriptor-object).

We should keep the `result` parameter as a JSON-Schema schema, and pass the `name` in as an optional parameter, ideally via an `options` object.

jlacivita (changes requested):

src/macrofier/engine.mjs
```diff
@@ -103,11 +105,12 @@ const getLinkForSchema = (schema, json) => {
   const dirs = config.createModuleDirectories
   const copySchemasIntoModules = config.copySchemasIntoModules
 
+  let name = schema.title || schema.name
   if (schema.schema) {
```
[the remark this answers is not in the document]
HaseenaSainul: yes removed. Now extended getLinkForSchema type with option {name: ''} to pass name also from the caller.

src/macrofier/engine.mjs
```diff
@@ -323,6 +326,7 @@ const generateMacros = (obj, templates, languages, options = {}) => {
   const imports = generateImports(obj, templates)
   const initialization = generateInitialization(obj, templates)
   const enums = generateEnums(obj, templates)
+  const enumsConversion = generateEnumsConversion(obj, templates)
```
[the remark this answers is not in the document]
HaseenaSainul: Sorry, it is created on top enum changes. Will rebase this branch once enum changes get merged, then it will sendup with only type related changes

src/macrofier/engine.mjs
HaseenaSainul: now updated logic to trim only there is only whitespaces, otherwise trimEnd of whitespace. It is added to avoided unwanted newlines

src/macrofier/engine.mjs
HaseenaSainul: it is creating types with unnamed. Changes was added to get proper names for all types generated

src/macrofier/engine.mjs
HaseenaSainul: updated signature with option name parameter and calling with proper name value

src/macrofier/engine.mjs
HaseenaSainul: removed

src/macrofier/engine.mjs
```diff
@@ -338,18 +342,19 @@ const generateMacros = (obj, templates, languages, options = {}) => {
   const events = eventsArray.length ? getTemplate('/sections/events', templates).replace(/\$\{event.list\}/g, eventsArray.map(m => m.body).join('\n')) : ''
   const eventList = eventsArray.map(m => makeEventName(m))
   const defaults = generateDefaults(obj, templates)
-  const schemasArray = generateSchemas(obj, templates, { baseUrl: '', section: 'schemas' }).filter(s => (options.copySchemasIntoModules || !s.uri))
+  const schemasArray = generateSchemas(obj, templates, { baseUrl: '', section: 'schemas' }).filter(s => (options.copySchemasIntoModules || !s.uri)).reverse()
```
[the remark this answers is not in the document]
jlacivita: Okay.

I'm fine if we keep this logic in C, but at some point, it might not work there either, since there's no guarantee that OpenRPC schema include schemas in any particular order.

src/macrofier/engine.mjs
jlacivita: got it, thanks.

src/macrofier/engine.mjs
```diff
@@ -1140,46 +1214,47 @@ function insertExampleMacros(template, examples, method, json, templates) {
...
+
+  const type = types.getSchemaType(result, json, { name: name, destination: state.destination, section: state.section  })
 
   if (result.type === 'object' && result.properties) {
     let content = getTemplate('/types/object', templates).split('\n')
 
     for (var i=0; i<content.length; i++) {
       if (content[i].indexOf("${property}") >= 0) {
-        content[i] = Object.entries(result.properties).map(([title, property]) => insertSchemaMacros(content[i], title, property, json)).join('\n')
+        content[i] = Object.entries(result.properties).map(([title, property]) => insertSchemaMacros(content[i], title || name, property, json)).join('\n')
```
jlacivita, line 1226: The `name` variable is the top-level name passed in via the method call.
The `title` variable is the individual title of each property the schema contains.

The `result.property` object will ALWAYS have name/value pairs, so the `title` variable above will always be populated. I think you can drop the `|| name`

src/macrofier/engine.mjs
```diff
@@ -1140,46 +1214,47 @@ function insertExampleMacros(template, examples, method, json, templates) {
...
+    const link = getLinkForSchema(result, json, { name: name})
 
     // if we get a real link use it
     if (link !== '#') {
       return `[${types.getSchemaType(result, json, { destination: state.destination, section: state.section  })}](${link})`
     }
     // otherwise this was a schema with no title, and we'll just copy it here
     else {
-      const schema = localizeDependencies(result, json)
+      const sch = localizeDependencies(result, json)
```
jlacivita, line 1244: Any reason for this shorter name?

src/macrofier/engine.mjs
```diff
@@ -1140,46 +1214,47 @@ function insertExampleMacros(template, examples, method, json, templates) {
...
-    return insertSchemaMacros(getTemplate('/types/default', templates), result.title, result, json)
+    return insertSchemaMacros(getTemplate('/types/default', templates), name, result, json)
   }
 }
 
 function insertSchemaMacros(template, title, schema, module) {
   return template.replace(/\$\{property\}/g, title)
-          .replace(/\$\{type\}/g, types.getSchemaType(schema, module, { destination: state.destination, section: state.section, code: false }))
-          .replace(/\$\{type.link\}/g, getLinkForSchema(schema, module))
+          .replace(/\$\{type\}/g, types.getSchemaType(schema.schema ? schema.schema : schema, module, { name: title, destination: state.destination, section: state.section, code: false }))
```
jlacivita, line 1256: `insertSchemaMacros` is only called by `generateResult`, which we have updated to always take a real schema as it's parameter, so I think these extra safety checks aren't needed now.

jlacivita (changes requested):

src/macrofier/engine.mjs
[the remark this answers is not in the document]
HaseenaSainul: This dependency already removed, now schemas sorting based on ReferencedBySchema

src/macrofier/index.mjs
```diff
@@ -200,7 +200,6 @@ const macrofy = async (
                 }
             })
 
-            console.log()
```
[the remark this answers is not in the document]
HaseenaSainul: already removed

src/macrofier/engine.mjs
HaseenaSainul: with previous changes, the variable schema was already used, hence tried a shorter name. But now that logic was changed and now only one schema variable so will change it back from sch to schema

src/macrofier/engine.mjs
HaseenaSainul: will update

src/macrofier/engine.mjs
HaseenaSainul: updated to use schema itself

src/macrofier/engine.mjs
```diff
@@ -1140,46 +1214,47 @@ function insertExampleMacros(template, examples, method, json, templates) {
...
+          .replace(/\$\{type.link\}/g, getLinkForSchema(schema.schema ? schema.schema : schema, module, { name: title }))
```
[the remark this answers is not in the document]

src/macrofier/engine.mjs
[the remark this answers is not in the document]
jlacivita: awesome!

PR #96, opened 2023-06-06
Static code implementation updates

HaseenaSainul:


jlacivita (approved):
Since this is all in the language pack, and outside of my area of expertise, I'm going to approve.

we should do a walk through of all these files before merging `c-language` into `next`

HaseenaSainul:
> Since this is all in the language pack, and outside of my area of expertise, I'm going to approve.
> 
> we should do a walk through of all these files before merging `c-language` into `next`

Sure

PR #97, opened 2023-06-06
Add support to generate file inclusion for common schema also

HaseenaSainul:


src/macrofier/engine.mjs
```diff
@@ -772,10 +795,14 @@ const generateImports = (json, templates) => {
     imports += getTemplate('/imports/x-method', templates)
   }
 
-  if (json['x-schemas'] && Object.keys(json['x-schemas']).length > 0) {
-    imports += Object.keys(json['x-schemas']).map(shared => getTemplate('/imports/default', templates).replace(/\$\{info.title\}/g, shared)).join('\n')
+  if (json['x-schemas'] && Object.keys(json['x-schemas']).length > 0 && !json.info['x-uri-titles']) {
```
jlacivita, line 798: I took a look at the code in `microfiber/index.mjs` and this `x-uri-titles` is being written, but never used.

What was you goal in using it here? maybe we can just remove it altogether from index.mjs & engine.mjs.

languages/c/templates/imports/common.c
```diff
@@ -0,0 +1 @@
+#include "JsonData_${info.title}.h"
```
jlacivita, line 1: common.c & default.c look the same now... is that intentional?

src/macrofier/engine.mjs
```diff
@@ -133,6 +134,24 @@ const getLinkForSchema = (schema, json, { name = '' } = {}) => {
   return '#'
 }
 
+const getComponentExternalSchema = (json) => {
```
jlacivita, line 137: You _might_ be able to use `getExternaSchemaPaths` from `shared/json-schema.mjs` instead of making a new method. Not sure if there's a minor difference, but worth considering.

src/macrofier/engine.mjs
```diff
@@ -29,7 +29,7 @@ import isString from 'crocks/core/isString.js'
 import predicates from 'crocks/predicates/index.js'
 const { isObject, isArray, propEq, pathSatisfies, propSatisfies } = predicates
 
-import { isRPCOnlyMethod, isProviderInterfaceMethod, getProviderInterface, getPayloadFromEvent, providerHasNoParameters, isTemporalSetMethod, hasMethodAttributes, getMethodAttributes, isEventMethodWithContext, getSemanticVersion, getSetterFor, getProvidedCapabilities, isPolymorphicPullMethod, hasPublicAPIs } from '../shared/modules.mjs'
+import { isRPCOnlyMethod, isProviderInterfaceMethod, getProviderInterface, getPayloadFromEvent, providerHasNoParameters, isTemporalSetMethod, hasMethodAttributes, getMethodAttributes, isEventMethodWithContext, getSemanticVersion, getSetterFor, getProvidedCapabilities, isPolymorphicPullMethod, isPolymorphicReducer, hasPublicAPIs } from '../shared/modules.mjs'
```
jlacivita, line 32: let's remove all polymorphic reducer stuff into it's own branch until we sort out the plan for that.

src/macrofier/engine.mjs
```diff
@@ -627,12 +646,11 @@ function generateSchemas(json, templates, options) {
 
   const schemas = (options.section.includes('methods') ? (hasMethodsSchema(json) ? json.methods : '') : (json.definitions || (json.components && json.components.schemas) || {}))
 
-  const generate = (name, schema, uri, { prefix = '' } = {}) => {
+  const generate = (name, schema, uri, { prefix = '', type = '' } = {}) => {
```
jlacivita, line 649: this is part of the `polymorphic-reduce` work right? if so, same comment.

src/macrofier/engine.mjs
```diff
@@ -693,13 +711,18 @@ function generateSchemas(json, templates, options) {
     }
     else if (schema.tags) {
       if (!isDeprecatedMethod(schema)) {
-        schema.params.forEach(param => {
-          if (param.schema && (param.schema.type === 'object')) {
-            list.push([param.name, param.schema, '', { prefix : schema.name}])
+        if (isPolymorphicReducer(schema)) {
```
jlacivita, line 714: same

jlacivita (changes requested):

src/macrofier/engine.mjs
```diff
@@ -345,7 +364,7 @@ const generateMacros = (obj, templates, languages, options = {}) => {
   const schemasArray = generateSchemas(obj, templates, { baseUrl: '', section: 'schemas' }).filter(s => (options.copySchemasIntoModules || !s.uri))
   const accessorsArray = generateSchemas(obj, templates, { baseUrl: '', section: 'accessors' }).filter(s => (options.copySchemasIntoModules || !s.uri))
   const schemas = schemasArray.length ? getTemplate('/sections/schemas', templates).replace(/\$\{schema.list\}/g, schemasArray.map(s => s.body).filter(body => body).join('\n')) : ''
-  const typesArray = schemasArray.filter(x => !x.enum)
+  const typesArray = schemasArray.length ? schemasArray.filter(x => !x.enum) : ''
```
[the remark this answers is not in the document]

languages/c/Types.mjs
```diff
@@ -167,6 +167,23 @@ const deepMergeAll = (module, name, schema, schemas, options) => {
   return union
 }
 
+const getPolymorphicReducerParamSchema = (method) => {
```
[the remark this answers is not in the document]
HaseenaSainul: Moved to new branch

src/macrofier/engine.mjs
HaseenaSainul: This was added to differentiate Common schema and module schema. I have found this is only available in common schema.

languages/c/templates/imports/common.c
HaseenaSainul: common.c is removed

src/macrofier/engine.mjs
HaseenaSainul: Actually this was added to get actual dependency of all common schema cases. Some common schema have dependency with common schema itself, say Discovery, Intents have dependency with Entertainment Schema and Entertainment has dependency with types, but as I mentioned above 'x-schemas' in common schema is not pointing its exact dependency. so this method is added to find out exact dependency of all components inside common schema.

src/macrofier/engine.mjs
HaseenaSainul: moved to separate branch

src/macrofier/engine.mjs
HaseenaSainul: yes, so moved to separate branch

src/macrofier/engine.mjs
HaseenaSainul: moved to separate branch

src/macrofier/engine.mjs
jlacivita: got it, makes sense.

jlacivita (approved):

PR #100, opened 2023-06-19
C-Language: Templates based property methods

sramani-metro:


src/macrofier/engine.mjs
```diff
@@ -1080,6 +1084,7 @@ function insertMethodMacros(template, methodObj, json, templates, examples={}) {
     .replace(/\$\{method\.params\.count}/g, methodObj.params ? methodObj.params.length : 0)
     .replace(/\$\{if\.params\}(.*?)\$\{end\.if\.params\}/gms, method.params.length ? '$1' : '')
     .replace(/\$\{if\.context\}(.*?)\$\{end\.if\.context\}/gms, event.params.length ? '$1' : '')
+    .replace(/\$\{method\.params\.instantiation\}/g,  paramInst)
```
jlacivita, line 1087: I think maybe `instantiation` isn't the best name for parameters (it works great for result, though)... we're actually instantiating the JSON versions of the params here, right?

Maybe this macro can be called something like one of these:

- `${method.params.json-values}`
- `${method.params.serialization}`
- `${method.params.transport}`

Thoughts?

jlacivita (changes requested):
Looks good in general, but I think we should pick a better name for the macro when it's used for parameters. See comment and let's discuss.

jlacivita (approved):

PR #103, opened 2023-06-22
Default template implementation added

HaseenaSainul:


src/macrofier/engine.mjs
```diff
@@ -175,12 +175,43 @@ const getMethods = compose(
...
 )
 
-const getSchemas = compose(
+const getComponentSchemas = compose(
   map(Object.entries), // Maybe Array<Array<key, value>>
   chain(safe(isObject)), // Maybe Object
   getPath(['components', 'schemas']) // Maybe any
 )
 
+const getSchemas = (schema) => {
```
sramani-metro, line 184: Instead of doing this for all the languages, can we do this based on a config option? namely anonymousSchemas which when set to true for a language, processes these method parameter types separately? We should leave the getSchemas as it was if the config is not set.

sramani-metro (changes requested):

src/macrofier/engine.mjs
HaseenaSainul: added config option to choose anonymous schema based on language

sramani-metro (commented):
looks good

sramani-metro (approved):

src/macrofier/engine.mjs
```diff
@@ -1129,7 +1132,7 @@ function insertMethodMacros(template, methodObj, json, templates, examples = {})
     .replace(/\$\{event\.callback\.params\.serialization\}/g, callbackSerializedParams)
     .replace(/\$\{event\.callback\.result\.instantiation\}/g, callbackResultInst)
     .replace(/\$\{event\.callback\.response\.instantiation\}/g, callbackResponseInst)
-    .replace(/\$\{info\.title\}/g, info.title)
```
jlacivita, line 1132: This will break several things. Why are you modifying `info.title` to be lowercase?

src/macrofier/engine.mjs
```diff
@@ -1109,9 +1109,12 @@ function insertMethodMacros(template, methodObj, json, templates, examples = {})
     .replace(/\$\{method\.params\.array\}/g, JSON.stringify(methodObj.params.map(p => p.name)))
     .replace(/\$\{method\.params\.count}/g, methodObj.params ? methodObj.params.length : 0)
     .replace(/\$\{if\.params\}(.*?)\$\{end\.if\.params\}/gms, method.params.length ? '$1' : '')
-    .replace(/\$\{if\.params.empty\}(.*?)\$\{end\.if\.params.empty\}/gms, method.params.length === 0 ? '$1' : '')
+    .replace(/\$\{if\.result\}(.*?)\$\{end\.if\.result\}/gms, resultType ? '$1' : '')
+    .replace(/\$\{if\.params\.empty\}(.*?)\$\{end\.if\.params\.empty\}/gms, method.params.length === 0 ? '$1' : '')
+    .replace(/\$\{if\.signature\.empty\}(.*?)\$\{end\.if\.signature\.empty\}/gms, (method.params.length === 0 && resultType === '') ? '$1' : '')
```
jlacivita, line 1114: How could a method have no signature?

jlacivita (changes requested):
We need to address the ${info.title} toLowerCase comment.

My other comments are questions about things I'd like to understand.

src/macrofier/engine.mjs
HaseenaSainul: Reverted this back. this was added to get module name in jsonrpc request in lower case. Since the ripple support both case, reverted it back

src/macrofier/engine.mjs
HaseenaSainul: if result of a method is, name:success and type:const/boolean. 
i.e, Discovery.signOut. Here result name is success and type is boolean, we don't want to consider this in function signature. similarly for type as const
if a method does not have any parameters and result as success + boolean/const:null. In such case the method signature become empty. C language we need to keep 'void' keyword for a function with out any parameter. Hence added signature.empty check to make sure appending this kind of function with 'void' keyword.

jlacivita: ah okay. yeah... that's a tough big of logic to name, but I get why you called it `signature.empty`

jlacivita (approved):

PR #106, opened 2023-06-26
SchemaSorting: logic updated to consider all schemas

HaseenaSainul:


src/macrofier/engine.mjs
```diff
@@ -673,6 +673,24 @@ function generateDefaults(json = {}, templates) {
...
 
+function sortSchemasByReference(schemas = []) {
+  let indexA = 0;
+  while (indexA < schemas.length) {
+
+    let swapped = false
+    for (let indexB = indexA + 1; indexB < schemas.length; ++indexB) {
+      const bInA = isDefinitionReferencedBySchema('#/components/schemas/' + schemas[indexB][0], schemas[indexA][1])
+      if ((isEnum(schemas[indexB][1]) && !isEnum(schemas[indexA][1])) || (bInA === true))  {
+        [schemas[indexA], schemas[indexB]] = [schemas[indexB], schemas[indexA]]
```
jlacivita, line 684: Woah, does that work? Impressive trick :)

jlacivita (approved):

src/macrofier/engine.mjs
HaseenaSainul: yes, it is sorting properly now :)

PR #109, opened 2023-07-07
Naming + void* to struct* changes based on sky review, removed redundant code, array handling fix

HaseenaSainul:


src/macrofier/engine.mjs
```diff
@@ -506,6 +506,8 @@ const insertMacros = (fContents = '', macros = {}) => {
   fContents = fContents.replace(/\$\{minor\}/g, macros.version.minor)
   fContents = fContents.replace(/\$\{patch\}/g, macros.version.patch)
   fContents = fContents.replace(/\$\{info\.title\}/g, macros.title)
+  fContents = fContents.replace(/\$\{module\.title\}/g, macros.title.toLowerCase())
```
jlacivita, line 509: Could we call this `${info.title.lowercase}` so it's clear that it is the same info as ${info.title} but in lower case form?

jlacivita (changes requested):
Minor macro rename change request.

src/macrofier/engine.mjs
HaseenaSainul: changed

jlacivita (approved):

PR #110, opened 2023-07-12
Generate polymorphci methods for anyOf param/result case

HaseenaSainul:


src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
     return json
 }
 
+const getRefDefinition = (uri = '', moduleJson = {}) => {
```
jlacivita, line 857: I think we already have a method in json-schema.json that does this. it's fine for modules.mjs to depend on json-schema, since all modules are json-schema files.

Can you try and leverage the existing methods for this? I think it's `getPath`

src/macrofier/engine.mjs
```diff
@@ -158,7 +158,7 @@ const getComponentExternalSchema = (json) => {
         let title = ''
         if (ref.includes('x-schemas')) {
           if (ref.split('/')[2] !== json.info.title) {
-            title = ref.split('/')[2]
+            title = ref.split('/')[2].toLowerCase()
```
jlacivita, line 161: It's hypothetically possible to have two unique schemas whose only different in title is case, e.g. `timezone` and `timeZone`.

We should never be doing this in the core codebase, aside from supporting a macro to output it into a template.

src/macrofier/engine.mjs
```diff
@@ -934,12 +934,30 @@ function generateExamples(json = {}, mainTemplates = {}, languages = {}) {
   return examples
 }
 
-function generateMethods(json = {}, examples = {}, templates = {}) {
+function generateMethods(json = {}, examples = {}, templates = {}, options = { polymorphic: false }) {
```
jlacivita, line 937: I think we should keep this function simple: it takes a list of method schemas and outputs the method bodies.

Let's move the code to insert additional methods higher up into generateMacros, right after we call `promoteAndNameSubSchemas`.

src/macrofier/engine.mjs
```diff
@@ -1113,6 +1131,7 @@ function insertMethodMacros(template, methodObj, json, templates, examples = {})
   template = insertExampleMacros(template, examples[methodObj.name] || [], methodObj, json, templates)
 
   template = template.replace(/\$\{method\.name\}/g, method.name)
+    .replace(/\$\{method\.json\.name\}/g, options.polymorphic ? methodObj.actualName || method.name : method.name)
```
jlacivita, line 1134: 1. let's rename the macro to `${method.rpc.name}`
2. I know we're just doing this in memory, but it' like to stick to valid OpenRPC. Can we leave the "actualName" value in `methodObj.name` and create a new `methodObj.title` field for the individual names, e.g. navigateToLaunchIntent.
3. we don't need to pass options around in order to do this... just change the code to:  `methodObj.title || methodObj.name` and it will work for all methods (plus we've created an OpenRPC compatible way to override method names in additional use cases if they come up, since any object can have a title property).

src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
...
+        }
+        else {
+            params.push(p)
+        }
+    })
+    return params
+}
+
+const generateResultAnyOfSchema = (method, methodResult, anyOf, anyOfTypes, title, summary) => {
+    let methodResultSchema = Object.assign({}, method.result.schema)
```
jlacivita, line 946: I'm not sure this does a deep copy... so you could be introducing side effects when you modify this object. Not sure if that's desired or not, but we try to avoid it in individual methods, especially in the /shared/ folder.

src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
...
+    else {
+      return definition
+    }
+  }
+  else {
+    throw `Cannot call getRefDefinition with ${uri}`
+  }
+}
+
+const getAnyOfSchema = (inType, json, anyOfTypes) => {
```
jlacivita, line 881: Why is `anyOfTypes` a parameter here? I noticed we are just passing in an empty array every time. Maybe it should be local to the `getAnyOfSchema` method body?

src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
...
+            if (inType.schema['$ref'] && (inType.schema['$ref'][0] === '#')) {
+                definition = getRefDefinition(inType.schema['$ref'], json, json['x-schemas'])
+            }
+            else {
+                definition = outType.schema
+            }
+            definition.anyOf.forEach(anyOf => {
+                anyOfTypes.push(anyOf)
+            })
+            outType.schema = anyOfTypes
```
jlacivita, line 897: a schema can't be an array... at least not in valid JSON-Schema.

src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
...
+        }
+    }
+    return outType
+}
+
+const generateAnyOfSchema = (anyOf, name, summary) => {
+    let anyOfType = []
```
jlacivita, line 904: This is initialized as an array here, but then properties are added to it later.
This is valid JavaScript, but it's pretty odd... the object will still be an array, and may trigger bad logic when parsing the schemas.

src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
...
+    if (!anyOf['$ref']) {
+        anyOf.name = anyOf.title || anyOf.name || ''
```
jlacivita, line 906: plain JSON schemas shouldn't have `name` attributes. Did you come across a situation where this was needed?

src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
...
+            methodParams.push(getAnyOfSchema(p, json, anyOfTypes))
+        }
+    })
+    if (isEventMethod(method)) {
+        methodResult.schema = getPayloadFromEvent(method)
+    }
+    methodResult = getAnyOfSchema(methodResult, json, anyOfTypes)
+
+    let polymorphicMethodSchemas = []
+    if (anyOfTypes.length > 0) {
```
jlacivita, line 988: Hmmm, I think we need separate logic for anyOf in the params vs the result.
Maybe you're handling it somehow, but it looks like you've got all anyOfs (from both params and results) in a single array, and don't know where they come from anymore...

src/shared/modules.mjs
```diff
@@ -854,6 +854,167 @@ const generateEventListenResponse = json => {
...
+            params: {},
+            result: {},
+            examples: {}
+        }
+        anyOfTypes.forEach(anyOf => {
+            let localized = localizeDependencies(anyOf, json)
+            let title = localized.title || localized.name || ''
+            let summary = localized.summary || localized.description || ''
+            polymorphicMethodSchema.actualName = method.name
+            polymorphicMethodSchema.name = `${method.name}With${title}`
```
jlacivita, line 1002: I was assuming that if the params have an anyOf, we'd use <method.name>With<title> for the new method, but if the result had an anyOf , we'd use `<method.name><Title>` since method names are generally verbs, you can say "navigateToHomeIntent" since HomeIntent is a result, but if you had a method called `walk(p: pet)` (and pet can be cat or dog) you would name them `walkWithDog` and `walkWithCat`

jlacivita (changes requested):
okay, there's some concrete change requests, and then I'm having trouble understanding some of the generateAllOf code...

Could you read my questions and we'll figure it out with more comments?

src/shared/modules.mjs
HaseenaSainul: Let me check and try

src/macrofier/engine.mjs
HaseenaSainul: Ok, will move

src/macrofier/engine.mjs
HaseenaSainul: Ok, let me try

src/macrofier/engine.mjs
HaseenaSainul: will revert, since this case conversion is already handling using info.title.lowercase

src/shared/modules.mjs
HaseenaSainul: No, this is added as an extra check to avoid

src/shared/modules.mjs
HaseenaSainul: yea, but I think if there is an anyOf case in both params and result that is not valid case. so there is only possibility to have this in one place. hence why used single variable to handle it

src/shared/modules.mjs
HaseenaSainul: Is there  a way in javascript to identify the string based on the value ?

src/macrofier/engine.mjs
HaseenaSainul: Here we have to use reverse like methodObj.name for individual names, e.g. navigateToLaunchIntent. and methodObj.title to keep actual name. since we are using method.name in all places, we will get new generated name properly for polymorphic case

src/shared/modules.mjs
HaseenaSainul: yes, this added to check the anyOf already defined, that means it is defined for params, and no need to proceed further parsing on this. like that added

src/shared/modules.mjs
HaseenaSainul: import { getPath as getRefDefinition } from './json-schema.mjs', since getPath is already importing from crocks/Maybe/getPath.js too

src/macrofier/engine.mjs

src/macrofier/engine.mjs
HaseenaSainul: moved generateMacros after handling promoteAndNameSubSchemas

src/shared/modules.mjs
HaseenaSainul: updated

src/shared/modules.mjs
HaseenaSainul: updated

src/shared/modules.mjs
HaseenaSainul: done

src/shared/modules.mjs
```diff
@@ -854,6 +855,137 @@ const generateEventListenResponse = json => {
     return json
 }
 
+const getAnyOfSchema = (inType, json) => {
```
jlacivita, line 858: Just making sure that I understand...

This method localizes the schema, and then reconstructs the anyOf $ref links.

It does this to ensure that we have a simple version of the schema, w/ links, instead of a more complicated one with complicate/multiple links?

src/shared/modules.mjs
```diff
@@ -854,6 +855,137 @@ const generateEventListenResponse = json => {
...
+            result: {},
+            examples: {}
+        }
+        anyOfTypes.forEach(anyOf => {
+
+            let localized = localizeDependencies(anyOf, json)
+            let title = localized.title || localized.name || ''
+            let summary = localized.summary || localized.description || ''
+            polymorphicMethodSchema.title = method.name
+            polymorphicMethodSchema.name = `${method.name}${title}`
```
jlacivita, line 973: Could we change this to:

```javascript
polymorphicMethodSchema.name = foundAnyOfParams ? `${method.name}With${title}` : `${method.name}${title}`
```
That way, users of our API will know that the method is being forked for parameters if the word "With" is in the name, and forked for results otherwise.

jlacivita (changes requested):
If we could just add the `With` to the method name, this looks good.

src/shared/modules.mjs
HaseenaSainul: yes, here we are populating content of the anyOf types reference. 
in case NavigationIntent, first we make it as localised, to get the solved $ref links. Then only we will able to identify it is anyOf or not (in case of $ref). Since localise resolve all subSchemas, we just check higher schema is $ref or not, if then get its definitions . i.e, 
"anyOf": [ { "$ref": "#/x-schemas/Intents/HomeIntent"},  
                  { "$ref": "#/x-schemas/Intents/LaunchIntent"},
                  { "$ref": "#/x-schemas/Intents/EntityIntent" },
                  { "$ref": "#/x-schemas/Intents/PlaybackIntent"},
                  { "$ref": "#/x-schemas/Intents/SearchIntent"},
                  { "$ref": "#/x-schemas/Intents/TuneIntent"},
                ]

src/shared/modules.mjs
HaseenaSainul: done

PR #118, opened 2023-07-24
detach setter declaration from property template and add setter templ…

HaseenaSainul:
…ate, remove unredundant warnings

### DIFF
--- a/src/macrofier/engine.mjs
+++ b/src/macrofier/engine.mjs
@@ -370,7 +370,7 @@ const addContentDescriptorSubSchema = (descriptor, prefix, obj) => {
   else {
     let descriptorName = capitalize(descriptor.name)
     let prefixName = capitalize(prefix)
-    title = (prefixName !== descriptorName) ? prefixName + descriptorName : descriptorName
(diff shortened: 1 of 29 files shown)


src/macrofier/engine.mjs
```diff
@@ -459,7 +459,7 @@ const generateMacros = (obj, templates, languages, options = {}) => {
   const allMethodsArray = generateMethods(obj, examples, templates)
   const methodsArray = allMethodsArray.filter(m => !m.event && (!options.hideExcluded || !m.excluded))
   const eventsArray = allMethodsArray.filter(m => m.event && (!options.hideExcluded || !m.excluded))
-  const declarationsArray = allMethodsArray.filter(m => m.declaration)
+  const declarationsArray = allMethodsArray.filter(m => m.declaration && (!options.hideExcluded || !m.excluded))
```
jlacivita, line 462: the excluded flag is sort of confusing... we only exclude it from the methods, not the declarations. It's for allowing a static template for a method that doesn't follow any of our templates.

We should remove this, or we'll break the JavaScript SDK.

HaseenaSainul: I have added this based on config option, since this creating redundant declaration for c code

PR #134, opened 2023-09-29
CPP SDK : changes to handle additional Properties

HaseenaSainul:


src/macrofier/types.mjs
```diff
@@ -137,6 +139,39 @@ const insertEnumMacros = (content, schema, module, name) => {
   return template.join('\n')
 }
 
+const insertObjectAdditionalPropertiesMacros = (content, schema, module, title, options) => {
+  options = options ? JSON.parse(JSON.stringify(options)) : {}
+
+  const options2 = JSON.parse(JSON.stringify(options))
+  options2.parent = title
```
jlacivita, line 146: You've already made a copy on line 143. Why do you need a second copy?

src/macrofier/types.mjs
```diff
@@ -137,6 +139,39 @@ const insertEnumMacros = (content, schema, module, name) => {
...
+  let jsonType = getJsonType(schema.additionalProperties, module)
+  if (!isPrimitiveType(jsonType)) {
+    jsonType = 'string'
+  }
+
+  const additionalType = getPrimitiveType(jsonType, 'additional-types')
+  let key = (propertyNames && propertyNames.title) ? propertyNames.title : getTemplate(path.join(options.templateDir, 'additionalPropertiesKey')).trimEnd()
+
+  content = content
+    .replace(/\$\{shape\}/g, shape)
```
jlacivita, line 163: Did you confirm that none of the JS templates were using `${property}` or `${Property}` here?

I forget why I had those two lines in there, but there must have been some reason.

src/macrofier/types.mjs
```diff
@@ -27,9 +27,11 @@ const primitives = {
   "integer": "number",
   "number": "number",
   "boolean": "boolean",
-  "string": "string"
+  "string": "string",
+  "float": "float"
```
jlacivita, line 31: The keys in the map are JSON-Schema primitives. JSON-schema does not have a `float` type, so it should not be here.

Did you come across a schema w/ a `float` type? I'm surprised it passed validation.

jlacivita: https://json-schema.org/understanding-json-schema/reference/numeric#:~:text=JSON%20does%20not%20have%20distinct,the%20same%20value%20in%20JSON.

jlacivita: These are just default values for JavaScript. If you need different overrides for Cpp, they can go in the sdk.config.json

src/macrofier/types.mjs
jlacivita: Also, I can't find where this new `insertObjectAdditionalProperitesMacros` method is called.

src/macrofier/types.mjs
```diff
@@ -151,21 +186,7 @@ const insertObjectMacros = (content, schema, module, title, property, options) =
...
-        .replace(/\$\{property\}/g, '')
-        .replace(/\$\{Property\}/g, '')
-        .replace(/\$\{shape\}/g, shape)
-        .replace(/\$\{parent\.title\}/g, title)
-        .replace(/\$\{title\}/g, type)
-        .replace(/\$\{delimiter\}(.*?)\$\{end.delimiter\}/g, '')
-        .replace(/\$\{if\.optional\}(.*?)\$\{end\.if\.optional\}/g, ''))
-    }
-  
+
```
jlacivita, line 189: Shouldn't we be calling  `insertObjectAdditionalProperitesMacros` here, with the same `if (schema.additionalProperties...` block?

src/macrofier/types.mjs
```diff
@@ -614,11 +635,17 @@ function getSchemaType(schema, module, { destination, templateDir = 'types', lin
     // }
   }
   else if (schema.type) {
-    // TODO: this assumes that when type is an array of types, that it's one other primative & 'null', which isn't necessarily true.
-    const schemaType = !Array.isArray(schema.type) ? schema.type : schema.type.find(t => t !== 'null')
-    const primitive = getPrimitiveType(schemaType, templateDir)
-    const type = allocatedProxy ? allocatedPrimitiveProxies[schemaType] || primitive : primitive
-    return wrap(type, code ? '`' : '')
+    const template = getTemplate(path.join(templateDir, 'additionalProperties'))
```
jlacivita, line 638: The current method (`getSchemaType`) doesn't recurse into each property... 

It looks like you're inserting the `Title` template if there happens to be a template for `additionalProperties`

I'm not sure I follow that. Could you explain?

src/macrofier/types.mjs
```diff
@@ -644,4 +685,4 @@ export default {
   getSchemaType,
   getJsonType,
```
jlacivita, line 686: getJSONType wasn't really being used before. I see that you're using it to do some local work in this method, which is cool, but we should stop exporting it now that you've got logic there.

Don't want any external dependencies on it.

jlacivita (changes requested):

src/macrofier/types.mjs
HaseenaSainul: additionalPorperties always seening like a map. FlatMap is additional property and it is map. Here we have to get only its title. FlatMap. hence added logic to get Title.

src/macrofier/types.mjs
HaseenaSainul: sorry missed to updated calling place. corrected now

src/macrofier/types.mjs
HaseenaSainul: I have not seen any use case from javascript. javascript generation not showed any issue

src/macrofier/types.mjs
HaseenaSainul: yea, it is already overridden language config. I missed to remove. initially I did not clear idea about this setting. finally I understood and missed to remove it

src/macrofier/types.mjs
HaseenaSainul: I updated code now to call insertObjectAdditionalProperitesMacros now

src/macrofier/types.mjs
HaseenaSainul: removed it

PR #136, opened 2023-09-29
CPPSDK: parameter and result handling changes

HaseenaSainul:


src/macrofier/types.mjs
```diff
@@ -359,14 +394,27 @@ function getSchemaShape(schema = {}, module = {}, { templateDir = 'types', name
     return insertSchemaMacros(result, schema, module, theTitle, parent, property)
   }
   else if (schema.anyOf || schema.oneOf) {
-    // borrow anyOf logic, note that schema is a copy, so we're not breaking it.
-    if (!schema.anyOf) {
-      schema.anyOf = schema.oneOf
+    const template = getTemplate(path.join(templateDir, 'anyOfSchema' + suffix))
```
jlacivita, line 397: Why do we need both an `anyOfSchema` template _and_ an `anyOf` template?

src/macrofier/types.mjs
```diff
@@ -479,8 +527,9 @@ function getSchemaType(schema, module, { destination, templateDir = 'types', lin
   if (schema['$ref']) {
     if (schema['$ref'][0] === '#') {
       const refSchema = getPath(schema['$ref'], module)
+      const refName = refSchema.title || schema['$ref'].split('/').pop()
       const includeNamespace = (module.info.title !== getXSchemaGroup(refSchema, module))
-      return getSchemaType(refSchema, module, {destination, templateDir, link, title, code, asPath, event, result, expandEnums, baseUrl, namespace:includeNamespace })// { link: link, code: code, destination })
+      return getSchemaType(refSchema, module, {destination, templateDir, link, title, code, asPath, event, result, expandEnums, baseUrl, namespace:includeNamespace, name:refName })// { link: link, code: code, destination })
```
jlacivita, line 532: I spent a lot of time removing the `name` param from `getSchemaType`. 

There's a method to generate names for every un-titled schema that runs before all of the code generation. Is it missing a use case?

src/macrofier/types.mjs
```diff
@@ -251,6 +282,7 @@ const insertTupleMacros = (content, schema, module, title, options) => {
   const itemsTemplate = getTemplate(path.join(options.templateDir, 'items'))
   const propIndent = (content.split('\n').find(line => line.includes("${properties}")) || '').match(/^\s+/) || [''][0]
   const itemsIndent = (content.split('\n').find(line => line.includes("${items}")) || '').match(/^\s+/) || [''][0]
+  const delimiter = getTemplate(path.join(options.templateDir, 'delimiter'))
```
jlacivita, line 285: `delimiter` is a pretty generic name. Maybe `tuple-delimiter` would be safer?

src/macrofier/engine.mjs
```diff
@@ -844,7 +844,7 @@ function generateSchemas(json, templates, options) {
     else {
       content = content.replace(/.*\$\{schema.seeAlso\}/, '')
     }
-//    content = content.trim().length ? content.trimEnd() : content.trim()
+    content = content.trim().length ? content : content.trim()
```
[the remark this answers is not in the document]
HaseenaSainul: Now I am trimming only if the it end up with whitespace alone. content.trim().length given zero if it is whitespace alone. So I think it will not make issue for any language

src/macrofier/types.mjs
HaseenaSainul: done

src/macrofier/types.mjs
HaseenaSainul: removed

src/macrofier/types.mjs
```diff
@@ -359,14 +394,27 @@ function getSchemaShape(schema = {}, module = {}, { templateDir = 'types', name
     return insertSchemaMacros(result, schema, module, theTitle, parent, property)
   }
   else if (schema.anyOf || schema.oneOf) {
-    // borrow anyOf logic, note that schema is a copy, so we're not breaking it.
-    if (!schema.anyOf) {
-      schema.anyOf = schema.oneOf
+    const template = getTemplate(path.join(templateDir, 'anyOfSchema' + suffix))
+    let shape
+    if (template) {
+      if (!template.includes('SKIP')) {
```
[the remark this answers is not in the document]
HaseenaSainul: SKIP is added to skip anyOf schema shape generation, since for CPP it is treating as string currently, so shape generation is not required. Updated template with comments

src/macrofier/types.mjs
HaseenaSainul: In CPP, shape and type has to be handled differently. so two templates are required. Currently anyOf is treating as string, so shape is not required. so adding one more two handle shape part

HaseenaSainul:
Regarding blackduck: these showing due to code match with template code std::string. could you please make this failure as ignore, since it is not code match

jlacivita (approved):

PR #138, opened 2023-10-06
CPPSDK: changes to handle enum and x-schema enum without title

HaseenaSainul:


src/macrofier/engine.mjs
```diff
@@ -686,7 +686,7 @@ const convertEnumTemplate = (schema, templateName, templates) => {
   const template = getTemplate(templateName, templates).split('\n')
   for (var i = 0; i < template.length; i++) {
     if (template[i].indexOf('${key}') >= 0) {
-      template[i] = enumSchema.enum.map(value => {
+      template[i] = enumSchema.enum.filter(value => value).map(value => {
```
jlacivita, line 689: Hmm, we currently don't have any enums w/ `false` or `null` in the values. we might want to map them to strings, though, e.g.: `false` -> `"false"` and `null` -> `"unknown"`.

`"unknown"` is our standard value for enums with an unknown possibility.

src/macrofier/types.mjs
```diff
@@ -120,17 +120,18 @@ const insertConstMacros = (content, schema, module, name) => {
...
-
-
-const insertEnumMacros = (content, schema, module, name) => {
+const insertEnumMacros = (content, schema, module, name, suffix) => {
   const template = content.split('\n')
 
   for (var i = 0; i < template.length; i++) {
     if (template[i].indexOf('${key}') >= 0) {
-      template[i] = schema.enum.map(value => {
+      template[i] = schema.enum.filter(value => value).map(value => {
```
jlacivita, line 128: Same comment here about false & null.

src/macrofier/types.mjs
```diff
@@ -120,17 +120,18 @@ const insertConstMacros = (content, schema, module, name) => {
...
         return template[i].replace(/\$\{key\}/g, safeName(value))
           .replace(/\$\{value\}/g, value)
       }).join('\n')
+      if (suffix !== ".cpp") {
```
jlacivita, line 132: Would be better to do this with a feature flag in sdk.config.json

jlacivita (changes requested):
I'm curious if you could write a summary of what this PR is doing?

We have an SDK.config.json feature to add titles to all untitled schemas. Why wouldn't we just use that, instead of this code?

src/macrofier/types.mjs
HaseenaSainul: ok

src/macrofier/engine.mjs
HaseenaSainul: ok, null is treating like empty line. hence why this change added. let me relook

HaseenaSainul:
> SDK.config.json

which one you meant here ? config.extractSubSchemas ? 
In latest accessibility.json schema we can see properties of Accessibility Schema such as FontFamily, FontEdge etc are changed to enums, but does not have title. these are x-schemas. So while searching for the x-schema property info titles are missing here and it is not generating types/shapes properly, even issue with namespace mapping.

src/macrofier/engine.mjs
HaseenaSainul: you can see

src/macrofier/types.mjs
HaseenaSainul: updated config for this

src/macrofier/engine.mjs
jlacivita: So these two fields need to actually support an unset value. we're using null for that in the JSON-Schema.
What makes sense in Cpp for an unset value?

We could return an error or create a standard enum value, like NONE.

If we filter it out, then the method won't support unset values.

HaseenaSainul: Will check about this

src/macrofier/types.mjs
```diff
@@ -157,15 +157,18 @@ const insertConstMacros = (content, schema, module, name) => {
...
   const template = content.split('\n')
 
   for (var i = 0; i < template.length; i++) {
     if (template[i].indexOf('${key}') >= 0) {
-      template[i] = schema.enum.map(value => {
+      template[i] = schema.enum.filter(value => value).map(value => {
         return template[i].replace(/\$\{key\}/g, safeName(value))
           .replace(/\$\{value\}/g, value)
       }).join('\n')
+      if (suffix !== enumSuffix) {
```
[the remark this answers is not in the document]
jlacivita: Then we can wrap the `,` in the template w/t that delimiter macro, and it won't get put on the last line.

JavaScript doesn't need the final delimiter, its happy with or without it.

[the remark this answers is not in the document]
HaseenaSainul: sure will check

src/macrofier/engine.mjs
HaseenaSainul: for cpp there is no Unset values, so just added to set value if there is unset template available. So for javascript this can be used to set unset values. So in this case, for CPP, there is chance to create enum with delimiter for last parameter as well (if last one is null), hence the delimiter handling logic is added after completing the value parsing

src/macrofier/types.mjs
HaseenaSainul: removed enumSufix from the config and used delimiter macro to handle this

src/macrofier/types.mjs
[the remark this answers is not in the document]
HaseenaSainul: removed enumSufix from the config and used delimiter macro to handle this

jlacivita (approved):
this looks good, thanks!

PR #140, opened 2023-10-10
CPPSDK: parameter/result handling for method and events + inner object handling

HaseenaSainul:
1. Handle parameters/result sections for method and callbacks based on templateDir passed
2. Handle Inner type inside object properties. For that introduced sub-property folder and keeps template for sub properties inside this folder. In CPP case need to handle sub property differently
Note: some templates already introduced in previous PRs anyOfSchemaShape/additionalProperties are already with these too, since this also comes under the parameter/result templates. will merge these only on top previous PR to ensure any change in these templates are in place


jlacivita (changes requested):
Okay this looks good, but I do believe that the 'generic' primitive is really just a default, and we should rename the template (and the code that loads it) to 'default'

jlacivita (approved):

PR #141, opened 2023-10-25
CPPSDK:

HaseenaSainul:
1. add overrideRule to handle template generation for object/primitives based on language requiredment
2. remove enum level checking inside getSchemaShape, to generate enums shape for all levels


src/macrofier/types.mjs
```diff
@@ -435,12 +441,12 @@ function getSchemaShape(schema = {}, module = {}, { templateDir = 'types', paren
...
 
   const suffix = destination && ('.' + destination.split('.').pop()) || ''
-  const theTitle = insertSchemaMacros(getTemplate(path.join(templateDir, 'title' + suffix)), schema, module, { name: schema.title, parent, property, required, recursive: false})
+  const theTitle = insertSchemaMacros(getTemplate(path.join(templateDir, 'title' + suffix)), schema, module, { name: schema.title, parent, property, required, recursive: false })
 
   let result = level === 0 ? getTemplate(path.join(templateDir, 'default' + suffix)) : '${shape}'
 
   let genericTemplate = getTemplate(path.join(templateDir, 'generic' + suffix))
-  if (enums && level === 0 && Array.isArray(schema.enum) && ((schema.type === "string") || (schema.type[0] === "string"))) {
+  if (enums && Array.isArray(schema.enum) && ((schema.type === "string") || (schema.type[0] === "string"))) {
```
jlacivita, line 449: Checking if level == 0 is important for enums in Typescript, because they can be defined inside of a type (similar to a cpp struct, but can have multiple levels)

I'm a bit concerned that you're taking it out here will break TypeScript declarations, but we won't see any issue in cpp, so you won't notice.

jlacivita (commented):
What exactly does `overrideRule` accomplish? I see you're using it in tuples and objects slightly different, but not exactly sure what the goal is?

jlacivita (approved):
These changes look good!

PR #143, opened 2023-11-01
fix: Errors thrown in async provider methods do not send the JSONRPC …

kpears201:
…error response


kevinshahfws (approved):

jlacivita (approved):
looks good to me.

PR #148, opened 2023-11-06
fix: Provider openrpc response and error methods have a wrapper object

kpears201:


src/macrofier/engine.mjs
```diff
@@ -1070,7 +1070,7 @@ function insertMethodMacros(template, methodObj, json, templates, examples = {})
   }
   const method = {
     name: methodObj.name,
-    params: methodObj.params.map(p => p.name).join(', '),
+    params: methodObj.params.map(p => p.name).join(''),
```
jlacivita, line 1073: This property is for providing a list of parameters, why'd you remove the comma? we want:

`param1, param2`

not 

`param1param2`

src/macrofier/engine.mjs
```diff
@@ -1093,7 +1093,7 @@ function insertMethodMacros(template, methodObj, json, templates, examples = {})
     }
   }
 
-  const paramDelimiter = config.operators ? config.operators.paramDelimiter : ', '
+  const paramDelimiter = config.operators ? config.operators.paramDelimiter : ''
```
jlacivita, line 1096: Similar issue here...

jlacivita (approved):

PR #159, opened 2023-12-15
CPPSDK: polymorphic method schema creation changes

HaseenaSainul:


src/macrofier/engine.mjs
```diff
@@ -1218,6 +1218,7 @@ function insertMethodMacros(template, methodObj, json, templates, examples = {})
   }
   const method = {
     name: methodObj.name,
+    title: methodObj.title,
```
jlacivita, line 1221: methods don't have a title property. what is this code doing?

jlacivita (changes requested):
This looks mostly good, but can we discuss the method.title comment?

src/macrofier/engine.mjs
HaseenaSainul: In https://github.com/rdkcentral/firebolt-openrpc/pull/159/files#diff-13e61cfe4b2442f6b926af45ded613df7d94c15f62d1ae9e5fd6508ba72c67b3R983 we are keeping actual name to title and we need to get actual name here to generate method name. But event case we need polymorphic.name created based on ResultAnyOf. Hence trying to use title of method

jlacivita: Okay, I see.

We can't have a macro in the template language called `${method.title}` because there's no such property in OpenRPC, and it's not clear what the difference between name & title is.

Can you change all of the macros back, and just add logic to ${method.name} to grab the correct value? Since our tooling's main purpose is code generation, we'll use the simple ${method.name} for the _code_ name.

If you need to also expose the original RPC method name as a macro (I think you do) then let's change. those macros to ${method.rpc.name}, we'll use the more specific ${method.rpc.name} for any place we want to generate the RPC method name.

HaseenaSainul: Ok, got it, will do like that

jlacivita (approved):

PR #160, opened 2023-12-15
CPPSDK: handle patternProperties

HaseenaSainul:


jlacivita (changes requested):
Hmm...

It's possible to have a schema with patternProperties AND explicit properties AND additionalProperties definitions...

I can't tell if your code handles this case, but it looks like maybe it does not.

HaseenaSainul:
> Hmm...
> 
> It's possible to have a schema with patternProperties AND explicit properties AND additionalProperties definitions...
> 
> I can't tell if your code handles this case, but it looks like maybe it does not.

In CPPSDK case for type, patternProperties always treating as string. But during the json parameter generation time we need to know to handle it like a single object instead string. Hence it handling only that part now.

jlacivita:
Okay, I'll make a test case to ensure we're doing it correctly before we merge to next.

Don't want to hold you up now.

jlacivita (approved):

PR #164, opened 2024-01-23
Check template for codeblocks exists or not before trying to generate

HaseenaSainul:
After adding this check able to reduce 17 minutes  from 25 minutes for Core SDK CPP generations and 2 minute from 3 minutes for Manage SDK CPP code generations


sramani-metro (approved):
LGTM

jlacivita (approved):
LGTM

PR #165, opened 2024-01-23
Optimization of generation time: generate macros only based on the schemhema template requirement if it is set

HaseenaSainul:
1. With the templateExtensionMap, able to reduce to 8 minutes from 17 minutes for Core SDK CPP generation and less than 2 minutes for Manage SDK CPP generation
2. Adding method type check for calling getSchemaShape inside insertMethodsMacros: able to reduce to 3 minutes from 8 minutes for Core, and few seconds reduced for Manage SDK case. 

As well as now JS generation is also reduced to less than 4 minutes from 7 minutes


jlacivita (approved):
This seems like a reasonable band-aid, but I'd like to understand why things got slow in the first place?

JS is not running that slow.

HaseenaSainul:
> This seems like a reasonable band-aid, but I'd like to understand why things got slow in the first place?
> 
> JS is not running that slow.

I have already explained the details through Slack. Hope that helps

PR #170, opened 2024-02-29
CPPSDK : Add Async communication support and CPP code generation for x-uses + x-allow-focus=true

HaseenaSainul:


src/shared/modules.mjs
```diff
@@ -289,6 +289,16 @@ const isPolymorphicReducer = compose(
     getPath(['tags'])
 )
 
+const isXUsesMethod = compose(
```
jlacivita, line 292: We should rename this `isAllowFocusMethod`.

Almost every method is an x-uses method, and this name really confused me when I was reading the code that calls it.

src/shared/modules.mjs
```diff
@@ -359,6 +369,8 @@ const getPublicEvents = compose(
 const hasPublicInterfaces = json => json.methods && json.methods.filter(m => m.tags && m.tags.find(t=>t['x-provides'])).length > 0
 const hasPublicAPIs = json => hasPublicInterfaces(json) || (json.methods && json.methods.filter( method => !method.tags.find(tag => tag.name === 'rpc-only')).length > 0)
 
+const hasXUsesMethods = json => json.methods && json.methods.filter(m => isXUsesMethod(m)).length > 0
```
jlacivita, line 372: same here.

src/macrofier/engine.mjs
```diff
@@ -108,7 +108,7 @@ const getTemplate = (name, templates) => {
 }
 
 const getTemplateTypeForMethod = (method, type, templates) => {
-  const name = method.tags && method.tags.map(tag => tag.name.split(":").shift()).find(tag => Object.keys(templates).find(name => name.startsWith(`/${type}/${tag}.`))) || 'default'
+  const name = method.tags ? (isXUsesMethod(method) && Object.keys(templates).find(name => name.startsWith(`/${type}/x-uses.`))) ? 'x-uses' : (method.tags.map(tag => tag.name.split(":").shift()).find(tag => Object.keys(templates).find(name => name.startsWith(`/${type}/${tag}.`)))) || 'default' : 'default'
```
jlacivita, line 111: We should rename this template `allowsFocus`

We don't include the `x-` that's just an OpenRPC convention. and see my other comments about allows focus vs uses.

jlacivita (changes requested):

src/shared/modules.mjs
HaseenaSainul: Sure, will update

src/shared/modules.mjs
HaseenaSainul: Will update

src/macrofier/engine.mjs
HaseenaSainul: will update

src/shared/modules.mjs
HaseenaSainul: done

src/shared/modules.mjs
HaseenaSainul: done

src/macrofier/engine.mjs

PR #188, opened 2024-05-10
App Pass-through

jlacivita:
This add support for the `x-provided-by` extension and `push-pull` method tag.


src/shared/modules.mjs
```diff
@@ -502,6 +615,60 @@ const createPullEventFromPush = (pusher, json) => {
...
+        paramsSchema.properties[p.name] = p.schema
+        if (p.required) {
+            paramsSchema.required.push(p.name)
+        }
+    })
+
+    return paramsSchema    
+}
+
+const createPullRequestor = (pusher, json) => {
```
kpears201, line 644: still want to keep this even though it is not used?

src/shared/modules.mjs
```diff
@@ -493,7 +550,63 @@ const createPullEventFromPush = (pusher, json) => {
...
+        if (t.name !== 'polymorphic-pull' && t.name)
+        {
+            event.tags.push(t)
+        }
+    })
+
+    return event
+}
+
+const createPullProvider = (pusher, params) => {
```
kpears201, line 562: I wonder if we should start moving these methods to their own file. modules.mjs is very large now.
Could just create a push-pull.mjs

jlacivita: I had the same thought. It basically serve two high level purposes:

- OpenRPC - OpenRPC generation
- Unified OpenRPC parsing / accessor methods

Might be worth splitting into two modules with those two purposes, as I often find myself wondering whether a given method operations on the pre-fireboltized documents or the post-fireboltized docuements.

Let's put on the backlog, as I've got a couple open branches I want to merge first and new engineers are getting involved that could take this refactor on.

kevinshahfws (approved):

PR #189, opened 2024-06-03
Feature/key name prefix

alkalinecoffee:
Adds a new `keyNamePrefix` value to include in our schema, allowing the developer to specify a custom prefix for enum keys.

For instance, an enum value `1080p60` would typically produce a key like `1080P_60`, but this would be disallowed as it starts with a number.

By specifying the `keyNamePrefix`, we can still use a mix of string and numerical values but can reference them with a valid stringified key.

```json
"VideoMode": {
  "enumKeyPrefix": "mode",
  "enum": [
    "1080p50",
    "1080p60",
    "unknown"
  ]
```

The above would produce enums such as:

```
MODE_1080P_50 = "1080p50"
MODE_1080P_60 = "1080p60"
MODE_UNKNOWN = "unknown
```

### DIFF
--- a/languages/c/src/types/NativeHelpers.mjs
+++ b/languages/c/src/types/NativeHelpers.mjs
@@ -60,7 +60,7 @@ const getNativeType = (json, fireboltString = false) => {
   }
   else if (jsonType === 'null' ) {
     type = 'void'
-  } 
+  }
   return type
 }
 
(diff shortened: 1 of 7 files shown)


alkalinecoffee:
switching to repo-local branch instead of personal branch

PR #231, opened 2024-11-18
fix: Make nested objects show in typescript .d.ts file and in the doc…

kpears201:
…umentation

### DIFF
--- a/src/macrofier/types.mjs
+++ b/src/macrofier/types.mjs
@@ -345,7 +345,10 @@ const insertObjectMacros = (content, schema, module, title, property, options) =
         }
         const schemaShape = indent + getSchemaShape(localizedProp, module, options2).replace(/\n/gms, '\n' + indent)
 
-        const type = localizedProp.type === 'object' ? getSchemaShape(localizedProp, module, { ...options2, type: true }): getSchemaType(localizedProp, module, options2)
+        let type = getSchemaType(localizedProp, module, options2)
+        if (type === 'object') {
+          type = getSchemaShape(localizedProp, module, { ...options2, type: true })
+        }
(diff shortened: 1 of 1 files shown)


kpears201:
Currently in both the documentation and the .d.ts file that ships with the SDK a type that has an inline-nested object looks like this:
```json
{
  "SectionIntent": {
    "description": "A Firebolt compliant representation of a user intention to navigate an app to a section not covered by `home`, `entity`, `player`, or `search`, and bring that app to the foreground if needed.",
    "title": "SectionIntent",
    "allOf": [
      {
        "$ref": "#/definitions/Intent"
      },
      {
        "$ref": "#/definitions/IntentProperties"
      },
      {
        "type": "object",
        "properties": {
          "action": {
            "const": "section"
          },
          "data": {
            "type": "object",
            "required": [
              "sectionName"
            ],
            "properties": {
              "sectionName": {
                "type": "string"
              }
            },
            "additionalProperties": false
          }
        },
        "required": [
          "data"
        ]
      }
    ]
  }
}
```
Has generated typescript like this:
```typescript
type SectionIntent = {
  action: 'section'
  data: object
  context: object
}
```

After this change it now looks like this:
```typescript
type SectionIntent = {
    action: 'section'
    data: {
      sectionName: string
    }
    context: {
      source: string
    }
  }
```

This allows developers looking at the documentation to actually know all the fields that are a part of this schema.
It also makes it so when developing a typescript app in an IDE, it can autocomplete all the fields instead of just the root object.


This will fix it for many schemas, most the intents have nested inline objects and have this issue.

kevinshahfws (dismissed):

ksentak (approved):

PR #234, opened 2024-12-02
Feature: connect on demand

tomasz-blasz:
Removes the need for the application to connect to the endpoint before using Firebolt, in other words, Firebolt will connect when necessary, on the first request. Also, if a disconnection occurs, Firebolt will try to reconnect when needed.

If the application stops using `Firebolt::IFireboltAccessor::Instance().Connect()`, it is not be notified about connection changes because connection change listener is provided in `Connect()`. To avoid that and allows the application to be notified and check status of connection, the following methods have been introduced:
- `RegisterConnectionChangeListener()` & `UnregisterConnectionChangeListener()`
- `IsConnected()`

Other fixes:
* If build step fails, build script exits with an error
* Log traces stop coloring on EOL
* `Firebolt::IFireboltAccessor::Instance()`: the instance in kept in a single variable: `_singleton`, additional one `instance` has been removed


tomasz-blasz:
already merged into `next` as `dde6753`

PR #243, opened 2025-02-17
fix: Dynamically detect openrpc file

AdityaKasar:
Previously json-engine was not able to use different openrpc file for different SDKs which has been fix here.


kschrief (approved):

tomasz-blasz:
merged into next in `a8363b0`

PR #250, opened 2025-03-14
Mock tests setup

AdityaKasar:


languages/cpp/src/shared/src/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ list(APPEND SOURCES
     Transport/Transport.cpp
     Accessor/Accessor.cpp
     Async/Async.cpp
+    Properties/Properties.cpp
```
tomasz-blasz, line 31: not needed, all methods in Propierties class are static and should remain static. Tests should verify code as it is

languages/cpp/templates/methods/property.cpp
```diff
@@ -6,8 +6,8 @@
         ${if.params}${method.params.serialization}${end.if.params}
         ${method.result.json} jsonResult;
 ${method.result.initialization}
-        ${if.params}Firebolt::Error status = FireboltSDK::Properties::Get(method, jsonParameters, jsonResult);${end.if.params}
```
tomasz-blasz, line 9: not needed, all methods in Properties are static

languages/cpp/src/shared/src/Gateway/Gateway.h
```diff
@@ -57,6 +57,11 @@ namespace FireboltSDK
 
         void TransportUpdated(Transport<WPEFramework::Core::JSON::IElement>* transport);
 
+        void UpdateGateway(std::unique_ptr<GatewayImpl> mockGateway)
```
tomasz-blasz, line 60: as it is only for tests, should be at least under #ifdef ... #endif block, not exposed for any other application using the API.

languages/cpp/src/shared/src/Gateway/unidi/gateway_impl.h
tomasz-blasz, line 1: Several things:
* Application class *MUST NOT* inherit from a class created for testing purposes. You should make tests for sources as they are without adopting them to your needs
* Keep existing formatting as it is, it makes comparison easier, I am referring to
  * Moving brackets to new lines
  * Changing argument convention from `type& arg` to `type &arg`, I am not arguing here which convention is better, just it is out of the scope of this task

languages/cpp/src/shared/src/Gateway/unidi/mockGateway.h
tomasz-blasz, line 1: That class cannot be a part of the application, it has been create for fulfill test scenarios, so its place is in test/ in SDK.

languages/cpp/src/shared/src/Gateway/unidi/gateway_impl.h
AdityaKasar: I agree for the formatting changes, I think my editors autoformator might have made that. I will try to revert it

languages/cpp/src/shared/src/IGateway.h
tomasz-blasz, line 1: That class cannot be a part of the application, it has been create for fulfill test scenarios, so its place is in test/ in SDK.

languages/cpp/src/shared/src/IProperties.h
tomasz-blasz, line 1: That class cannot be a part of the application, it has been create for fulfill test scenarios, so its place is in test/ in SDK.

languages/cpp/src/shared/src/Properties/IPropertiesMock.h
tomasz-blasz, line 1: That class cannot be a part of the application, it has been create for fulfill test scenarios, so its place is in test/ in SDK.

languages/cpp/src/shared/src/Properties/Properties.cpp
tomasz-blasz, line 1: Please, do not modify existing sources to fulfil test requirements.

languages/cpp/src/shared/src/Properties/Properties.h
tomasz-blasz, line 1: Please, do not modify existing sources to make tests working.
Also, the same as for Gateway, application class *MUST NOT* inherit from mock classes created just for testing purposes.

languages/cpp/templates/methods/property.cpp
tomasz-blasz, line 1: If Properties class won't be changes to make tests, that change would be not  needed.

tomasz-blasz (changes requested):
Overall, mock classes for testing purposes cannot exist at application level *AND* application classes cannot inherit from them.

PR #266, opened 2025-05-27
feat: Stripping all null value optional params

vladimir-rangelov:


languages/javascript/src/shared/Transport/index.mjs
```diff
@@ -103,6 +103,32 @@ export default class Transport {
...
+      let i = keys.length - 1
+      while (i >= 0) {
+        const key = keys[i]
+        if (params[key] === null) {
+          delete params[key]
+          console.warn('WARNING: null values for optional params will be disallowed in a future Firebolt version. Parameter: ' + key)
+        } else if (params[key] === undefined) {
+          // undefined means the param is not provided, which is Ok
+        } else {
+          break
```
tomasz-blasz: This `break` is causing that some scenarios are not handled properly, e.g.
```js
// for a method signature with 1 required and 2 optionals args:
p(a1 /* required */, a2 /* optional */, a3 /* optional */)

// and the following call:
p(1, null, 2);
```
the `null` at the second position is not removed:
```js
let p = {
  "a": 1,
  "b": null,
  "c": 2
};
let numOfOptionalParams = 2;

console.log(`before : p: ${JSON.stringify(p)}, numOfOptionalParams: ${numOfOptionalParams}`);
removeNullOptionalParams(p, numOfOptionalParams);
console.log(`after  : p: ${JSON.stringify(p)}`);

// > before : p: {"a":1,"b":null,"c":2}, numOfOptionalParams: 2
// > after  : p: {"a":1,"b":null,"c":2}
```

languages/javascript/src/shared/Transport/index.mjs
```diff
@@ -103,6 +103,32 @@ export default class Transport {
...
+        }
+        i--
+        if (numOfOptionalParams > 0) {
```
tomasz-blasz: There is something wrong with range checking
```js
// having a method signature with one required and one optional arguments:
p(a1 /* required */, a2 /* optional */)

// and the following call:
p(null, null);
```
`null` at both positions are removed, while the first one should stay:
```js
let p = {
  "a": null,
  "b": null
};
let numOfOptionalParams = 1;

console.log(`before : p: ${JSON.stringify(p)}, numOfOptionalParams: ${numOfOptionalParams}`);
removeNullOptionalParams(p, numOfOptionalParams);
console.log(`after  : p: ${JSON.stringify(p)}`);

// > before : p: {"a":null,"b":null}, numOfOptionalParams: 1
// > WARNING: null values for optional params will be disallowed in a future Firebolt version. Parameter: b
// > WARNING: null values for optional params will be disallowed in a future Firebolt version. Parameter: a
// > after  : p: {}
```

languages/javascript/src/shared/Transport/index.mjs
```diff
@@ -103,6 +103,32 @@ export default class Transport {
     return Transport.get()._sendAndGetId(module, method, params, transforms)
   }
 
+  static removeNullOptionalParams(params, numOfOptionalParams) {
+    if (numOfOptionalParams > 0) {
+      // iterate over all params starting backwrods and if the param is null or undefined remove it from the params object
+      const keys = Object.keys(params)
+      let i = keys.length - 1
+      while (i >= 0) {
```
tomasz-blasz: Understanding of the logic inside the `while` is not straightforward, mostly because of 2 `break` statements and decremental of 2 variables. I think the function could be simplified to something like that:
```js
function removeNullOptionalParams(params, numOfOptionalParams) {
  if (numOfOptionalParams > 0) {
    const keys = Object.keys(params)
    const len = keys.length;
    for (let i = len; i > Math.max(len - numOfOptionalParams, 0); i--) {
      const key = keys[i-1]
      if (params[key] === null) {
        delete params[key]
        console.warn('WARNING: null values for optional params will be disallowed in a future Firebolt version. Parameter: ' + key)
      }
    }
  }
  return params
}
```
It also fixes issues mentioned below.

tomasz-blasz (changes requested):
There are 2 issues in `removeNullOptionalParams()` and the implementation of the function could be simplified.

languages/javascript/src/shared/Transport/index.mjs
vladimir-rangelov: I refactored the function and added more comments

languages/javascript/src/shared/Transport/index.mjs
vladimir-rangelov: yes, I added a fix and a unit test

languages/javascript/src/shared/Transport/index.mjs
vladimir-rangelov: If we remove the middle argument and show the client of the API a warning "Do not pass null for optional arguments", then the client is at dead end, because it cannot provide "c":2 without providing  "b". I think we should stop removing null optional arguments the moment we reach an argument with a value

tomasz-blasz: Makes sense.

tomasz-blasz (dismissed):

tomasz-blasz (approved):

PR #267, opened 2025-05-29
Stripping null values for optional params in bidirectional

vladimir-rangelov:


languages/javascript/templates/sdk/Gateway/index.mjs
```diff
@@ -17,10 +17,40 @@
  */
 
 
+function removeNullOptionalParams(params, numOfOptionalParams) {
```
tomasz-blasz, line 20: Please, fix indentation in this function.

tomasz-blasz (changes requested):

languages/javascript/templates/sdk/Gateway/index.mjs
vladimir-rangelov: Fixed

tomasz-blasz (approved):

PR #272, opened 2025-07-04
fix: Treating enums as a subschema breaks the creation of the entityI…

vladimir-rangelov:
…nfo type


tomasz-blasz (approved):
_just for information_: the change is fine for  JS and MD, but it brakes C++ generated code (`std::optional<'prod' | 'test'>` and so on). If we would like to fix C++ code at some point, we can use that fix only for JS code, but when C++ is generated, then fallback to existing implementation.

PR #274, opened 2025-09-05
Fix events and subscribers in the JS generated documentation

vladimir-rangelov:


src/macrofier/engine.mjs
```diff
@@ -88,6 +88,39 @@ const setConfig = (c) => {
   config = c
 }
 
+const isGeneratingDocs = (languages) => {
+
+  if (languages && Object.keys(languages).some(lang => lang.toLowerCase() === 'json-rpc'))
```
tomasz-blasz, line 93: minor: as in functions below, you could just simplify it into one-liner 'return' statement.

tomasz-blasz (approved):

src/macrofier/engine.mjs
vladimir-rangelov: Yes, it started with more stuff in it, but then got reduced. We will refactor this code at some point, for now I'll leave it as is.

PR #275, opened 2025-09-11
fix: Omit private methods from doc ToC

alkalinecoffee:
The documentation portal has an issue where HTML tags are being escaped/not properly handled by Jekyll.  I don't think this is an issue with Jekyll, but rather in how we're mixing markdown with HTML in a non-standard way (this wasn't tested against the Comcast doc portal yet).

This mainly affects the private methods/events, which were intended to be "hidden" behind a dropdown in the Table of Contents.e.g.:

Screenshot 2025-09-11 at 12 02 28 PM

This changes updates the logic to remove the hidden methods/broken HTML, and instead replace it with a link to the private methods/events section.

Screenshot 2025-09-11 at 12 02 43 PM

### DIFF
--- a/languages/markdown/templates/sections/private-events.md
+++ b/languages/markdown/templates/sections/private-events.md
@@ -1,6 +1,3 @@
 ## Private Events
-<details id="private-events-details">
-  <summary>View</summary>
 
-  ${event.list}
-</details>
+${event.list}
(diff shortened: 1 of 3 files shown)


tomasz-blasz (approved):
Looks fine to me, but I'd like @vladimir-rangelov to review it before we merge.

vladimir-rangelov (approved):

PR #279, opened 2025-10-07
Fix regressions in generated documentation

vladimir-rangelov:
Completely rewrite removeUnusedSchemas and add unit tests for it.


src/macrofier/index.mjs
```diff
@@ -270,32 +271,61 @@ const macrofy = async (
...
+                    Object.assign(externalSchemas[id].components.schemas, schemas)
+                }
+        })
+        
+        // update the refs
+        Object.values(externalSchemas).forEach( document => {
+            getLocalSchemas(document).forEach((path) => {
+                const parts = path.split('/')
+                // Drop the grouping path element, since we've pulled this schema out into it's own document
+                if (parts.length === 4 && path.startsWith('#/x-schemas/' + document.info.title + '/')) {
```
tomasz-blasz: please, comment on why there is a 4 — or even better, extract it as a const. It’ll make our lives much happier when inspecting the code later 😉
_applies to line 299 as well_

tomasz-blasz (commented):
all good apart from mysterious '4'

src/macrofier/index.mjs
vladimir-rangelov: Fixed.

tomasz-blasz (approved):

